PR #25, opened 2017-11-25
  add Arch/Manjaro instruction and some fix for this distributives

beliys:
1) add Arch/Manjaro instalation instruction to README.md
2) fix xfreerdp synthax (https://aur.archlinux.org/cgit/aur.git/tree/0000-fix-xfreerdp-synthax.patch?h=asbru-cm-git)
3) fix crash getXWindowsList (https://aur.archlinux.org/cgit/aur.git/tree/fix-pacmanager-crash-getXWindowsList.patch?h=asbru-cm-git)
4) fix travis setup


gfrenoy:
Thanks a lot for your help and great contribution Alexander :)

1) README --> ok
2) .travis.yml --> we actually do not use it and should be removed
3) xfreerdp --> ok
4) crash getXWindowsList --> commenting everything out is a quick workaround but we should find the root reason, shouldn't we ?  any idea why it crashes ?

gfrenoy:
Alexander, since you now use "/parent-window" to start a freedrp session, I don't believe that we need to start that time (that is trying to find the external window which obviously do not exist) so I would suggest to remove the condition

    $$self{_CFG}{environments}{$$self{_UUID}}{method} eq 'RDP (xfreerdp)'

at line 1074.

beliys:
Now travis.yml removed from pull request.
Im only ported original patches from https://aur.archlinux.org/packages/pacmanager-git/
About crash getXWindowsList - we have open issues in pac manager repository - https://github.com/perseo22/pacmanager/issues/25

gfrenoy:
Ok I see, thanks for the reference !

I cannot reproduce the problem but I don't really get the point of that "_EMBED_KIDNAP" part of the code.  For sure it is now useless for RDP.  I made some test with or without for VNC but really did not see any difference... Maybe we should just get rid of it ?

figue:
Hey, I'm Arch's maintainer in the AUR.
Thanks for these fixes.
Beware that these lines in lib/pac_conn:

```
+			$$CFG{'tmp'}{'width'}-=5;
+			$$CFG{'tmp'}{'height'}-=5;
```
were put to avoid an odd calculation of the parent window with some GTK themes. Sometimes xfreerdp is bigger that the parent window and server has scrollbars... Very annoying. That's entirely added because I can't figure out how do the right math to get a proper width/height. I'm sure you guys are better in Perl than me ;) In fact, I saw that in most of the GTK themes, substracting 5 pixels is enought, but it's completely arbitrary.

Thanks also to maintain this fork.

beliys:
What is need for merging patches now? 
Need save only README.md and fix xfreerdp synthax and create issue for getXWindowsList?

gfrenoy:
Yes it would be better to think a bit more on the "getXWindowsList" issue.  I started to look at it be could not find time to finalize.

I was planning to merge your changes and then apply some more change on the getXWindowsList.  I believe these will prevent the crash...

I you want to apply the README and xfreerdp quicker, can you uncomment the "getXWindowsList" issue ?

Thanks.

@figue  Thanks for your additional comment Joan, I was indeed a bit surprised by the hard coded values.  I believe it's good enough for now.

gfrenoy:
Thanks Alexander !  Also for creating the issue #27 :)  No time this week to dig into this but it is definitely the worth to have a closer look to the root issue.

PR #26, opened 2017-11-27
[Usability] Add tooltip to "Edit" -> "Expect" when disabled

merikz:
When Authentication is set to Manual in the "Edit Connection" dialogue, the "Expect" tab is disabled.
As a newcomer I didn't realize that it was the Manual setting that disabled it, I just wanted Expect automation even if the password was not stored in the connection settings.
In fact the meaning of "Manual" means that all automation is turned off, which is perfectly sensible when you think about it.

This patch makes the connection between Manual authentication and the disabled expect tab visible to the user by not only changing the "sensitive" setting of the tab (greying it out), but also setting the tool-tip of the tab when authentication mode is changed.

### DIFF
--- a/lib/PACEdit.pm
+++ b/lib/PACEdit.pm
@@ -211,10 +211,21 @@ sub _initGUI {
 
 sub __checkRBAuth {
 	my $self = shift;
-	
+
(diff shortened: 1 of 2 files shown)


gfrenoy:
Cool !  Thanks :)

Issue #28, opened 2017-11-28
User Variables does not work


merikz:
The "User Variables" that can be set per connection does not expand correctly when used. 
For example, in an "Expect" or "Remote Macro" you can use this as "Send" or "Command":
```
echo uservar0=<V:0>
```
The result when running the expect or command is:
```
user@host:~$ echo uservar0=HASH(0x2f15b18)
-bash: syntax error near unexpected token `('
```
Assuming the value of User Variable 0 is "Var0Value" you'd expect this result:
```
user@host:~$ echo uservar0=Var0Value
uservar0=Var0Value
```

This was earlier reported at https://sourceforge.net/p/pacmanager/bugs/249
I have a fix for this. PR coming.

PR #29, opened 2017-11-29
[Bugfix] - User variables doesn't work

merikz:
This fixes Issue #28.
Technically the problem was that: 
  - The {txt} hash key was not referenced when replacing user variables

As the User Variables have been broken in PAC since forever it seems, I guess people do not use it.
Here's how I use it:
 Create a template connection, with values as <V:0>, <V:1> wherever unique values are needed, such as ip-address, jump hosts, proxies and user name. Then just copy the connection and change all the connection unique values on the User Variables tab, instead of hunting down the different places that needs to be changed.

  https://sourceforge.net/p/pacmanager/bugs/249/
  Issue https://github.com/asbru-cm/asbru-cm/issues/28

### DIFF
--- a/lib/PACConfig.pm
+++ b/lib/PACConfig.pm
@@ -297,7 +297,7 @@ sub _setupCallbacks {
 			# Populate with user defined variables
 			my @variables_menu;
 			my $i = 0;
-			foreach my $value ( @{ $$self{variables} } ) {
+			foreach my $value ( map{ $_->{txt} // '' } @{ $$self{variables} } ) {
 				my $j = $i;
 				push( @variables_menu, {
 					label => "<V:$j> ($value)",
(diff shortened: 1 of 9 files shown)


gfrenoy:
Thanks a lot Magnus.  Maybe we should have a PACUtils function for the transformation ?  Anyway, it should work.

And indeed, must not be much used.  I remember I tried a couple of weeks ago but could not make it working so I gave up :-/  Probably a page in the wiki explaining the purpose and giving an example would help...

merikz:
There are a lot of features hidden in Asbru, that you only find after much exploration. An introduction for newcomers on the wiki might be a good thing? Problem is that I'm not a newcomer anymore, so I don't remember the difficulties. Maybe add a "help wanted" on the issue list? 
IMHO tool-tips could be used to help users get going, if done right.

Anyway, I have an extension of the User Variables coming up, merging User Variables and Global Variables so User Variables also gets names instead of numbers. I made it 3 years ago, but this was a bit more work to port to asbru. PR in the works.

PR #53, opened 2018-01-17
Don't swallow key-accelerators which are not used by asbru

nomike:
I've revisited the section of code in PACTerminal.pm handling key-combinations in a terminal window.
asbru-cm now only swallows key-strokes it really uses and passes the others to the underlying terminal widget.

E.g. there is a plugin called "dircycle" for oh-my-zsh which allows cycling through recently visited directories with <ctrl>+<shift>+<leftarrow>/<rightarrow>. This was not possible prior to this patch.

While I was already there I also reformatted the code a little bit (indentations, newlines, etc.) to make it more readable and more consistent.


gfrenoy:
Special thanks for reformatting :-)

PR #64, opened 2018-02-20
Removed deprecated "have" function

daugustin:
The "have" function in bash-completion is deprecated. Indeed, using this results in completions not working.

See the [Gentoo Wiki](https://wiki.gentoo.org/wiki/Bash/Installing_completion_files) for example.


gfrenoy:
Thanks :)

PR #69, opened 2018-03-31
Change PAC Manager references to Àsbrù Connection Manager

GreenItSolutions:
Hi,
Thanks a lot to continue PAC MAnager Project.
We can't help you to develop but we can test and help you to improve Àsbrù

Just some cosmetic changes in this PR.

### DIFF
--- a/lib/PACConfig.pm
+++ b/lib/PACConfig.pm
@@ -65,7 +65,8 @@ my $RES_DIR			= $RealBin . '/res';
 # Connect to Gnome's GConf
 my $GCONF			= Gnome2::GConf::Client -> get_default;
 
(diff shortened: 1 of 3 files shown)


gfrenoy:
Thanks for your help and collaboration :)  PR are not that frequent but always very much appreciated.
But in that case I do fear this change is a bit more complicated than what you propose :-/

Changing the cipher will make existing encrypted data unreadable.  So as an example, if a GUI password is set, one won't be able to start the GUI.

The way to go would probably be something like:

- make sure we only have one place where the cipher is defined
- define a new cipher for Àsbrú
- detect if we are importing a legacy PAC configuration file or a new Ásbrú configuration
- Use the correct cipher according to the configuration version

GreenItSolutions:
ok, i understand the problems. But my knowledge is too small to make the changes.
I prefer close this PR and hope you add these changes in future releases.

Issue #71, opened 2018-03-31
Split horizontal/vertical reversed


gfrenoy:
'Split horizontally with TAB' will create a vertical split, and vice versa.

See also https://sourceforge.net/p/pacmanager/bugs/197/

gfrenoy:
@nugzarg thanks, this one was easy and the fix is ready for the next release.

PR #73, opened 2018-04-02
Usability fixes: split window, append group name semantics

basos9:
Two fixes:

    The split window semantics. In the present version the split semantics are swapped (vertically splits horizontally). This is fixed.

    The "automatically append group name to tittle for new connections" semantics. In the present version the group name is prefixed. This is fixed.

### DIFF
--- a/lib/PACTerminal.pm
+++ b/lib/PACTerminal.pm
@@ -2364,6 +2364,9 @@ sub _split {
 	my $uuid_tmp	= shift;
 	my $vertical	= shift // '0';
 	
+	# NOTE: For some reason the vertical flag is negated
+	$vertical = ! $vertical;
+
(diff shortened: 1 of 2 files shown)


KlaasT:
Thanks for your contribution. Pull Requests are always very welcome.

It would be great if you could sign our CLA. Just click on the CLA button or this.

basos9:
I will correct the email and re post.
Thanks

gfrenoy:
Thanks @basos9 ; please also consider #71 since I think you fixed the same issue.

basos9:
Hello,
This seems to fix the same, with different way.

Also I have another small fix for the connection group.

I reopened #74, that fixed my email.

Thanks

PR #74, opened 2018-04-02
Usability fixes: split window, append group name semantics (corrected)

basos9:
Two fixes:

 - The split window semantics. In the present version the split semantics are swapped (vertically splits horizontally). This is fixed.

- The "automatically append group name to tittle for new connections" semantics. In the present version the group name is prefixed. This is fixed.

### DIFF
--- a/lib/PACTerminal.pm
+++ b/lib/PACTerminal.pm
@@ -2364,6 +2364,9 @@ sub _split {
 	my $uuid_tmp	= shift;
 	my $vertical	= shift // '0';
 	
+	# NOTE: For some reason the vertical flag is negated
+	$vertical = ! $vertical;
+
 	my $tabs = $self -> {_NOTEBOOK};
 	$$self{_SPLIT_VERTICAL} = $vertical;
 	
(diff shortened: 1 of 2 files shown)


gfrenoy:
Thanks a lot !

About this one:

> The split window semantics

Do we really need this commit ?  I think https://github.com/asbru-cm/asbru-cm/commit/f613e1786012ddef55c5f3b6ca9896367a9b588d is already fixing the issue, isn't it ?

gfrenoy:
About the 2nd proposal:

>  In the present version the group name is prefixed

Well, yes, this was "as designed" ; why do you want to change this ?

basos9:
The 2nd:
The setting in the preferences states automatically **append** group name to tittle for new connections.
Despite that the current behavior is to prepend instead of append
i.e.  fixed
hostname - group name
i.e. current
group name - hostname


I fixed it because I think  that it is more usable, when you have many open tabs to view the important part. When many tabs are open you tend to view only the first part of the tab title. Which when you have many tabs of the same group, is the same for all.

gfrenoy:
Oh ok I got it now !  Thanks for the additional explanation :)

Would you mind reverting https://github.com/asbru-cm/asbru-cm/pull/74/commits/3fd90602db188d779185ef67296cb29e0aff8c47 (1st point) since it has already been fixed ?  Then I'll merge the 2nd change.

Thanks a lot for your help :)

PR #77, opened 2018-04-04
Fix append group name semantics (append instead of prepend)

basos9:


gfrenoy:
Wonderful, thanks a lot !

jserras:
Could this be made into an option, prepend, append or not there, i preferred the other the other way.

gfrenoy:
I believe it could, another PR to manage this will be fore sure welcome ;)

A tip for the one (many ;-p) who would like to give it a try : look for "append group name" in the code.

jserras:
I already saw and tracked the relevant code, but I'm not a dev, anyway I think I can manage given enough time, know enough perl and git, never messed with glade, but I'll manage.

PR #86, opened 2018-05-19
Tweaks for more vertical terminal space

lukefor:
- Added option to auto hide main button bar when on terminal tabs
- Reduced height of macro buttons 

Rationale behind this is basically trying to make Asbru look identical to SecureCRT on Windows, where the UI consists of tabs, terminal and macro bar only, with everything else hidden behind menus. IMO there's no need for one-click access to preferences, about, etc. while a terminal is in use


gfrenoy:
Very cool !  Thanks a lot.

Is there any option foreseen to restore back the main button bar once hidden ?

gfrenoy:
Let me merge the change (btw: thanks!) and open a new issue for the additional request.

PR #103, opened 2018-08-05
Add option to change main window title with terminal title.

kak-tus:
Sometimes we need to change window title. As example, to allow keepassx insert suitable password by main window title.


kak-tus:
I used PAC with this fix for 4 years. Because of author did not maintain PAC.
Now I know about this fork, and I will be happy if this fix will be in master.

gfrenoy:
Thanks !

PR #124, opened 2018-12-14
update split panes separator handling

syco:
Hello,
I changed how the separator in split panels is set and managed.

1. On split panels creation the separator is now set in the middle, instead of being closer to the left side.
2. There's a new menu item to reposition the separator in the middle after the window is resized.
3. There's also a new option in the global settings to force the separator to stay in the middle when the window is resized.

There's one thing I'm not sure about:
```perl
$$self{_SPLIT_VPANE} -> set_position( ( ( $vertical ? $y : $x ) / 2 ) - 7 );
$$self{_SPLIT_VPANE} -> set_position( ( ( $$self{_SPLIT_VERTICAL} ? $y : $x ) / 2 ) - 7 );
```
that " - 7" is there to compensate the width of the separator itself, so that when I do "tput cols" on both terminals they return the same number of columns.
But I guess that on other os, or with a different gtk theme, the separator could be thinner/thicker.
Do you think there's any way to put the separator exactly in the middle and have 2 terminals with the same number of columns?
Thanks


gfrenoy:
Thanks for this improvement ; it's indeed way much better.

Don't know how to find out the size of the separator ; I guess it won't make a huge difference so it's good enough for now (until we can get a Gtk specialist on board, I guess).

PR #136, opened 2019-03-06
Added instruction for Mint 19 (tara) and Mint 19.1 (tessa) to README.md

wiserweb:
Added instruction for Mint 19 (tara) and Mint 19.1 (tessa) to README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,15 @@ To do so, execute the following commands:
   $ sudo apt-get install asbru-cm
   ````
 
+- Mint 19 (tara), 19.1 (tessa)
+
+  ````
+  wget https://packagecloud.io/install/repositories/asbru-cm/asbru-cm/script.deb.sh
+  chmod +x script.deb.sh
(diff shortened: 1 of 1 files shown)


KlaasT:
Thanks for your contribution. 

However we should put this on hold for now. For the moment it is possible to install the latest snapshot  on Mint 19 without going through Ubuntu Bionic. As soon as the new version is released this should also work for the stable repository.

gfrenoy:
The following should now work for both Tara and Tessa:

```
curl -s https://packagecloud.io/install/repositories/asbru-cm/asbru-cm/script.deb.sh | sudo bash
curl -s https://packagecloud.io/install/repositories/asbru-cm/snapshots/script.deb.sh | sudo bash
sudo apt install asbru-cm
```

As soon as the new release is out, the "snapshot" trick is not even needed so what's currently displayed on the homepage is OK.

gfrenoy:
With the latest release, current instructions are correct and shall not be changed.

PR #162, opened 2019-08-02
GUI: down 'Set keyboard focus' button to bottombox to gain a piece of…

figue:
… screen in embed tabs


gfrenoy:
Great change ! Thanks a lot.

PR #163, opened 2019-08-09
Improving RDP methods

figue:
This PR will merge my latest patches on RDP methods. Quick changelog:

- Both methods have now a new free text box to add any option not implemented in Ásbrú
- Fixed rdesktop width size when X11:GUITest is used (it adds 2 pixels in width, don't know why)
- Added "-K: keep window manager key bindings" as a new button in rdesktop (very useful)
- Removed rdesktop percent size, it's no longer supported upstream, so I left only: fullscreen, embed and fixed size
- Fixed some typos and improved some tooltips
- Fixed percent size in xfreerdp, regex was wrong

As always, feel free to comment and request any explanations you need... I will be on vacation a couple of weeks, so maybe if a Perl guru is reading this can review and improve the code. I tried to be correct and think as simple as I can to implement the sub _parseCfgToOptions. This function now is a big "if/else" because I need to capture all parameters in one hand, and in another hand the rest of the options (those that comes from the Other Options text box). Specially in rdesktop, regex was tricky to find, because you need to catch individual options (such as -f -z) and options with one argument (such as -a 32, or -r clipboard:CLIPBOARD).

### DIFF
--- a/lib/PACTerminal.pm
+++ b/lib/PACTerminal.pm
@@ -546,13 +546,6 @@ sub _initGUI {
 
 				$$self{_GUI}{_SOCKET} = Gtk2::Socket -> new;
 				$sc2 -> add_with_viewport( $$self{_GUI}{_SOCKET} );
-
-			$$self{_GUI}{_BTNFOCUS} = Gtk2::Button -> new_with_mnemonic( 'Set _keyboard focus' );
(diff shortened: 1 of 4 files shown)


figue:
Conflicts must be some commits that were already pushed a week ago... I think it could be easily resolved.

Issue #173, opened 2019-09-16
RSA authentication on MOSH protocol


bitchecker:
**Describe the bug**
Using mosh protocol, is not possible to use RSA authentication.

**To Reproduce**
1. Create a new connection using MOSH protocol
2. Try to use Private Key authentication
3. See that Private Key is disabled ( not clickable )

**Expected behavior**
Setup rsa

**Desktop (please complete the following information):**
 - OS: Fedora
 - Version: 30

**Additional context**
I had open ad issue on PAC connection manager ( https://github.com/perseo22/pacmanager/issues/74 ) forked and send a Pull Request ( https://github.com/perseo22/pacmanager/pull/75 ) i think that the same code can be used here.

gfrenoy:
PR #174 has been merged.  Let us know if it works for you so we can close this issue.

Thanks again for your help and collaboration.

bitchecker:
@gfrenoy It works properly...I have tested it before create pull request.

It's a pleasure to help when you can, also because I use this connection manager( also protocol and auth type ) in life and at work ( I'm a Unix SysAdmin! lol )

PR #174, opened 2019-09-16
Enabled RSA Authentication via MOSH protocol

bitchecker:
fix for https://github.com/asbru-cm/asbru-cm/issues/173


gfrenoy:
Looks great, thanks for your contribution !
(note: I cannot test it so let me know if it works for you)

Issue #205, opened 2019-10-10
[gtk3] Clusters sync input no longer works


hanspr:
If I open 2 terminals that are in a cluster, and I type in one of them the other one does not receives a key but apparently a reference to a HASH. It used to work in Gtk2 version

![imagen](https://user-images.githubusercontent.com/1572396/66614696-a0ac2e00-eb8f-11e9-866e-97272c8e4158.png)

But if I use the  Power Cluster controler, the keys are sent correctly to both terminals

![imagen](https://user-images.githubusercontent.com/1572396/66614713-b3266780-eb8f-11e9-956d-50b2a1eda746.png)

May be I find the solution and send it later.

Linux Mint

hanspr:
I found a some solutions to all my reported issues they are on a pull request. Hope it helps

gfrenoy:
Thanks !  I cherry-picker your changes from your branch.

Issue #204, opened 2019-10-10
[gtk3] Can not type passprhasse, wrong terminal focus, can not login to any server


hanspr:
My terminal configuration

![imagen](https://user-images.githubusercontent.com/1572396/66609192-05f72380-eb7e-11e9-8520-5fb374f33a08.png)

1.- Double clic to open a session
2.- Window will show up with no focus

![imagen](https://user-images.githubusercontent.com/1572396/66609298-48b8fb80-eb7e-11e9-8d59-09f4c77e2555.png)

If I type the passphrasse it will show instead of being hidden

![imagen](https://user-images.githubusercontent.com/1572396/66609322-58d0db00-eb7e-11e9-929d-47401ec13141.png)

After typing the the connection is lost, no passphrasse ever transmitted

You can only connect if your key has no passphrase

**Desktop (please complete the following information):**
 - Linux Mint
 - 18.3

hanspr:
More tests

Fresh install, no configurations changed, only added one connection by hand and on connection, it opens a new tab, no focus, no passphrase entered

![imagen](https://user-images.githubusercontent.com/1572396/66610456-368c8c80-eb81-11e9-9af3-52b59e277604.png)

If you keep watting it ends with a disconnect error

![imagen](https://user-images.githubusercontent.com/1572396/66610487-499f5c80-eb81-11e9-8c02-fabdd12408bd.png)

hanspr:
I manged to make it work with this regex

([pP]ass|[pP]ass[wW]or[dt](\s+for\s+|\w+@\w+)*|[cC]ontrase.a|Enter passphrase for key ).*:.*$

But now it asks me for my passphrase with a popup window

![imagen](https://user-images.githubusercontent.com/1572396/66611988-d4825600-eb85-11e9-8d90-d80b95b88d0a.png)

hanspr:
Managed to fix this for me, I don't know why are you forcing this window on publickeys?

Removed ask for password on public keys

![imagen](https://user-images.githubusercontent.com/1572396/66612686-3c39a080-eb88-11e9-9209-358e07771561.png)

Or I'm missing a new configuration option?

hanspr:
The only remaining problem is focuse again

hanspr:
Managed to fix focus once and for all too.

Added this line.

![imagen](https://user-images.githubusercontent.com/1572396/66614337-4d85ab80-eb8e-11e9-9748-f54bc01a9485.png)

[...]

Issue #203, opened 2019-10-10
[gtk3]  Asbru crashes if you inactivate option "Put Terminals y Tabs"


hanspr:
**Describe the bug**
In a fresh new installation.
Open Asbru, creates config dir

Open Preferences

Inactivate option : Put terminals in new tabs

![imagen](https://user-images.githubusercontent.com/1572396/66600683-a8f27200-eb6b-11e9-8f46-e9093802b450.png)

Save preferences, close, reopen and following message appeares

![imagen](https://user-images.githubusercontent.com/1572396/66600707-b871bb00-eb6b-11e9-8090-0578d7c44d11.png)

**Linux Mint:**
 - Linux
 - Version Xenial

hanspr:
The consequence is that in PACMain.pm this next validation forbids hiding main PAC Windows after opening a connection on a new window

Line 2768

( $$self{_CFG}{'defaults'}{'hide on connect'} && ! $$self{_CFG}{'defaults'}{'tabs in main window'} ) and $self -> _hideConnectionsList;

egmontkob:
Thanks for testing the GTK3 branch!

PACMain.pm line 835 should be `Gtk3::Window::set_default_icon_from_file( $APPICON );`. I fixed it somewhere else in the code, but not here.

(Not sure how line 2768 is relevant, I didn't run into a problem there.)

hanspr:
Thanks for the fix egmontkob,

I can confirm that it does not crash any more.

The relevace of line 2768, is that if you can not uncheck this option (tabs in main tab, because produces the crash) then you are not able to use option "hide on connect", because they are connected by an "&&"

I use windows, not tabs and like to hide the Main Asbru window after I double click on the connection. I had to comment that if, and removed the and condition, to be able to do use hide on open, because I could not uncheck tabs in main tab option.

I had removed my commented line and returned line 2768 to its original validation and everything works fine.

Thanks again.

hanspr:
Sorry

gfrenoy:
Thanks both for your help !  The fix has been applied on branch gtk3.

PR #213, opened 2019-10-11
Gtk3 Fixes and proposed improvements

hanspr:
Bear with me, I'm really new to git, I hope I'm doing things correctly.

Summary of this merge:

- Fix to issue #203 , suggested by egmontkob
- Fix to issue #204 , change to regex expect, removed force to type passphrase if  password is not `<<ASK>>`, added grab focus event to return focus to newly opened terminal
- Removed temporarily right click, it conflicts with new terminal applications. Should be moved to shift right click when I learn how to do it
- Added tiling new windows when opened in cluster, examples

Open 4 windows
imagen

Open 6 windows
imagen

Mi editor removes trailing white spaces so created some very large differences, sorry.


hanspr:
I have added more changes to fix issue #205 , is more a workaround I guess, I submit it so you can review and maybe get an Idea how to make cluster keys work again. I don't, know maybe is my vte library that is not so new (version 0.42). Hope it helps

gfrenoy:
Hans, thanks **a lot** for your help !  I really appreciate it and will eventually include your suggestions to the main code.

But...

To make things easier to merge:

- Can you avoid changing "white spaces" (or have a separate commit for only cosmetic changes)
- Create one PR per issue (or at least reduce the number of issues per PR...)
- Identify the fixed issue in the commit
- Avoid leaving commented lines (dead code)
- Only leave comment that explain the code (not who or when it has been done, we have git for that)

Thanks again for your great contribution and collaboration !

hanspr:
I'm sorry, is the editor I always have it set up to remove trailing spaces, when I edit code from here I will turn it off.

I will remove the commented code, old habits from another era, will try to catch up

PR #223, opened 2019-10-15
Gtk3format

hanspr:
This is the good one.

No changes to global variables yet, but the code is ready to implement.


KlaasT:
Suppose #221 can be ignored as the coding standards file is included in this one. From a first review I can't find any breaking changes here.

hanspr:
I still do not know how to send a single file, when I add a second pull request it brings both.

So as I understand I should have waited to the first pool request to be closed to send the second one alone.

PR #225, opened 2019-10-17
Gtk3format PACMain

hanspr:
Coding Standard applied to PACMain.pm

Documentation Added to PACMain.pm

Checked before delivery
* Cut, Copy, Paste of nodes
* Configuration changes applied in runtime, saved files correctly
* Opening Sessions
* Opening Cluster Session
* Fresh first run creates files correctly
* General revision of clicks on all gui interfaces
* Start, Quit, no warning, errors.
* Tested with Tabs
    * New connections
    * Change tabs
    * Exit connections
    * Openning cluster with tabs
* Cluster GUI
    * Listing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,3 +68,5 @@ after_deploy:
   - pip install -r ./packpack/tools/requirements.txt
   - python ./packpack/tools/packagecloud prune ${PACKAGECLOUD_USER}/snapshots deb ${OS} ${DIST} --keep 5
   - python ./packpack/tools/packagecloud prune ${PACKAGECLOUD_USER}/snapshots rpm ${OS} ${DIST} --keep 5
+  - python ./packpack/tools/packagecloud prune ${PACKAGECLOUD_USER}/gtk3 deb ${OS} ${DIST} --keep 5
+  - python ./packpack/tools/packagecloud prune ${PACKAGECLOUD_USER}/gtk3 rpm ${OS} ${DIST} --keep 5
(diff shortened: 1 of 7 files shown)


KlaasT:
This is mainly formatting cleanup, did I understand that correctly?

hanspr:
Hi @KlaasT 

This is formatting and reimplementing implicit ifs and unless statements, one liners to the coding standard. When I finished this pull request is when I realized that I need it to split them in 2, to make it safer in the future.

Though I did my testing to make sure that there are no broken parts, it is convenient that other people does the same or at least begins using this version to see if something comes up. We all use the terminal options differently and that makes it hard for one person to check all possible combinations. All changes on this code are related to the main Asbru window, global, settings, the nodes tree interaction, Tabs, and the how terminals are opened : on a Window or a Tab.

In the next files, I will divide the pull requests in two, one for the formatting (which can be safely applied because the changes will be only cosmetic)

And the second one with the reimplementation of : ifs, one liners, etc. To make the code easier to follow.

Trying to learn as fast as I can.

KlaasT:
Splitting between the Tasks is a good Idea. From my point of View don't worry too much in making mistakes. For this reason we are working in a branche and not the master.

gfrenoy:
Thanks for your help!  I looked at the new look of ```PACMain.pm``` and it definitely looks better :-)

I'm thinking about the best way to merge this to the ```gtk3``` branch since as it is today, it'll be quite difficult to isolate the various changes.

But maybe as Klaas suggested, we should not worry too much and simply test as much as we can on the "unstable" (gtk3) branch.

Did not have much time this week but I'll have a closer look this weekend.

Thanks again !

PR #228, opened 2019-10-18
Gtk3format PACTerminal.pm

hanspr:
This pull request is for PACTerminal.pm

This are cosmetic changes only: tabs to spaces, removal of excessive spaces and indentation correction.

Should be safe to merge.


hanspr:
Well I still do not understand correctly this but here is the pull request with the full coding standard added. removed : unless, reversed ifs, implicit ifs, remove one liners.

PR #235, opened 2019-10-20
Gtk3format. PACUtils.pm coding standard

hanspr:
PACUtils.pm with coding standard, formatting.

Had to fix a bug on popup menus when group or connection names have "<>' characters in it is a minor fix and it creates no conflict with existing menu names.

PACTerminal.pm

Fixed a regression showing names on tabs or menus when contain characters : "<>'


gfrenoy:
The new "look" of the code is really great !  Much appreciated you clean this up :)

PR #236, opened 2019-10-20
Gtk3format Large cosmetic change

hanspr:
This cosmetic changes, no coding standard to the rest of the files.

PACTerminal.pm and PACUtils.pm are included again, I guess I send the wrong commit before or have made more changes afterwards, really don't know. Still understanding this process.

From here one I will try to fix a new file every one or 2 weeks, the most important have been covered.


gfrenoy:
Thanks again !

From now on, maybe you should start from a new branch of gtk3... Btw, if you did not read it yet, this article is interesting to better understand the concepts : https://jeffkreeftmeijer.com/git-flow/.

hanspr:
I read a similar article or the git-flow page some while back, the problem is that until now all was theory no real life experience, so I still make mistakes in the order of the commands. I will drop the gtk3format and come back to the gtk3 one again.

Thanks

gfrenoy:
For a first time, you're doing it very good (I can't believe I'm writing this ;-p).  Thank again for your help and collaboration, it is really more comfortable for me to read the code today so it really helps !

PR #238, opened 2019-10-21
Gtk3 Add global configuration settings

hanspr:
### Added global configuration settings

This change allows to setup a global configuration settings in asbru-cm and not having to rewrite  code in all related modules that need a reference to the configuration directory.

This will ease the future move of `.config/pac` to `.config/asbru`

Also added command line option : `--configdir=path`

This allows to have a default .config/pac directory with our own production settings for our day to day work.

And start asbru-cm with a temporary different configuration path for testing. Where we can add, delete, rename and test connections without messing our production settings or having to rename folders in ./config directory.

Examples of use:

use ~/.config/pac.test
```bash
asbru-cm --configdir=pac.test
```

use /home/user/Documents/pac
```bash
asbru-cm --configdir=/home/user/Documents/pac
```

moved priority of execution of migration function before creating any new directories, so eventually configuration directory can by migrated with a simple directory rename:

`mv ~/.config/pac ~/.config/asbru`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,3 +68,5 @@ after_deploy:
   - pip install -r ./packpack/tools/requirements.txt
   - python ./packpack/tools/packagecloud prune ${PACKAGECLOUD_USER}/snapshots deb ${OS} ${DIST} --keep 5
   - python ./packpack/tools/packagecloud prune ${PACKAGECLOUD_USER}/snapshots rpm ${OS} ${DIST} --keep 5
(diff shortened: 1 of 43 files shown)


gfrenoy:
Great enhancement !  Thanks a lot :-)  :1st_place_medal: 

One very minor remark, by looking at all the existing options, shouldn't we have ```--config-dir``` instead ?

hanspr:
done

PR #241, opened 2019-10-22
Gtk3 changed configdir to config-dir

hanspr:
Changed name of command line option configdir to config-dir


hanspr:
Fixed issue #242

imagen

PR #244, opened 2019-10-23
Gtk3 Save sessions popup menu

hanspr:
Mouse right on terminal : Save Session.

Would not save session on selected directory

Would save on .config/pac/tmp/... all the time

Clean escape sequences selection would not clean saved session correctly

This fixes both issues


KlaasT:
Nice find. I could imagine this should or could be backported to master. Is anyone against backporting this to the old gtk2 version?

PR #249, opened 2019-10-26
Fix #168

### LINKED ISSUE #168: Cannot interact with multiple selections in Connection list
**Describe the bug**
Multi selections in Connections list become deselected.

**To Reproduce**
1. Highlight multiple connections. Either via CTRL+click or SHIFT+click
2. Right click on any of the highlighted ones.
3. They all get deselected, except for the one you right clicked on.

**Expected behavior**
I'd expect to be able to highlight multiple connections and use some menu items that could apply to more than 1 connection at a time. Such as the "Export SELECTED Connections...", "Start", "Delete".

**Screenshots**

 - OS: Kubuntu 18.04
 - Version: Asbru CM 5.0

hanspr:
Fix to issue #168 

Enhancement: Added plural to some words when there are more than one node to copy

After click on "Copy nodes", all nodes get deselected now.

Fixed a second bug : Paste with keys : inserts multiple nodes as expected, but paste with right click would paste only the first node. This is fixed too.

Paste option is insensitive, when there is no data on $$self{_COPY}{'data'}, it was always available even if there was nothing to paste. (hope there is no side effect, haven't noticed jet)


KlaasT:
One additional question. Did you by any chance rebase your fork to the origin HEAD? For example (this is an example for master):

```
git remote add upstream https://github.com/asbru-cm/asbru-cm.git
git fetch upstream
git checkout master
git reset --hard upstream/master  
git push origin master --force 
```

Be aware if you do that now you would lose your local changes.

PR #255, opened 2019-10-31
fix copy with mouse selection gets lost sometimes

hanspr:
Fix annoying bug where clipboard gets lost after copy using the mouse selection on terminal

The problem is the function gets called when text is selected and deselected, if you select text and keep it selected, you can shift insert. But if the focus on the selected text gets lost, then the clipboard gets a blank text and you loose your clipboard data.

Now the clipboard is persistent as long as no new selection is made.


KlaasT:
lol. I had that issue and I thought it had something to do with my Windows VM and the shared clipboard being broken.

PR #257, opened 2019-11-01
Fix mouse on cluster

hanspr:
On relation to my previous fix on #205 

The previous solution fixed sync keys on cluster terminals, but sync mouse events was lost.

The correct solution was to reimplement the calls on vteFeedChild on PACUtils.pm

Now the keys and mouse work in a cluster

Keys sync correctly, again

imagen

And mouse events sync correctly on cluster terminals, so you can interact simultaneously with the mouse on cluster terminals again

imagen


hanspr:
Sorry with this one, it has all the previous commits together again, so I guess I made a mistake one more time with this git thingy

So cherry peek, to get the changes on : PACTerminal.pm and PACUtils.pm

PR #258, opened 2019-11-01
Add pcc tray menu

hanspr:
Adds PCC to tray menu for quick access

Fixed a small regression with last PR, _cluterCommit, has to be outside "record command history" option.

Guess this one will accumulate with previous ones to.


KlaasT:
Thanks for that PR. We seem to have some conflicts here. If you could have a look at them that would be really great.

hanspr:
Done

Issue #262, opened 2019-11-05
gtk3 Minimalist Main Window


hanspr:
@gfrenoy

Hi, Gaëtan:

I would like to see if you (or any from the group) could help me with this implementation in case you consider it would be good for the project.

I know, must people like tabs, I'm the total opposite to that, I like windows. So, for window users, the Main Asbru Window is large and takes a lot of unnecessary space.

I have implemented this layout, that is very small and I would like to show it as close to the tray icon when it shows. It has all the necessary functions I need and the rest can be easily accessible with right click or the menu from the tray (like preferences)

![imagen](https://user-images.githubusercontent.com/1572396/68176738-a9eca880-ff4b-11e9-98d8-70b7c73311cb.png)

The help I need is to edit the asbru.glade and create a new option like "presentation minimal" (in the  "Look and Feel" tab), or if you think the concept could be grown to have 2 or more available layouts. Then add a combo.

![imagen](https://user-images.githubusercontent.com/1572396/68177048-b6bdcc00-ff4c-11e9-9693-2a55a8de794a.png)

Al least we could have a: "Traditional" or "Tabbed" and "Minimal" layout.

With this option I could insert the necessary "ifs" to have both layouts available. And if there is a request for another layout there would be the option to try to implement it.

Once the option is there I would implement the saving, reading and handle the change. I guess it could be possible to implement it without a need to restart, is basically show and hide buttons and attach to one or other container.

Thanks.

gfrenoy:
Interesting :)  I personally never really get used to that "GIMP"-like layout but I perfectly understand this can be a very good way of working.

But as there are already a lot of options, this might probably add even more confusion so we might need to think a bit if we cannot merge some of these options together...

Especially that "Put Terminal TABS in Main Window" section might be the same as the "Tabbed Layout", isn't it ?

hanspr:
> Especially that "Put Terminal TABS in Main Window" section might be the same as the "Tabbed Layout", isn't it ?

Not really, there is a difference.

"Put Terminal TABS in Main Window" instructs asbru do what you think

![imagen](https://user-images.githubusercontent.com/1572396/68232217-3a65d000-ffc2-11e9-93b0-568512c1c842.png)

but if you uncheck it

The terminals begin to tab to the first terminal created or to an external tab window.

![imagen](https://user-images.githubusercontent.com/1572396/68232283-65502400-ffc2-11e9-9697-85abb67885e3.png)


[...]

PR #264, opened 2019-11-07
Add layouts proposal

hanspr:
In reference to enhancement proposal #262 

I have finished the option to have a Traditional (current layout on the Main Ásbrú Window), or a Compact Main Ásbrú Window

View in Ubuntu gnome-shell, with tabbed terminals in external window

imagen

View in Mint cinnamon

imagen

Option added in Behaviour

imagen

While testing in Ubuntu and checking the warnings in the configuration for, decided to add a routine at asbru-cm to check for the running environment. It is not fully tested, but the must important environment to detect is gnome-shell

With this environment variable, actions can be taken to control the interactions of Asbru, limitations or features can be implemented for that particular desktop.

In this case the Compact layout in gnome-shell

* disables tray icon and closing to tray so the window is not lost.
* adds Settings to the click-right on the main window to have access to settings because there is no tray icon

During the coding process some small warnings about transient windows were fixed as well


gfrenoy:
Thanks @hanspr :)  Could not find time to give it a try but I'll do it asap (at latest this WE) !

hanspr:
If you are going to test, I would recommend that you activate this options:

* Main options > Behaviour > Hide on connection = checked
* Terminal > Look and feel > Open new connections : test with both Tabs and Windows
* Terminal > Look and feel > Show status bar on terminal = not checked

The idea is that the main window pops up, you open a terminal, it hides and you have a terminal (tabbed or alone) to work.

You need a second one. Activate on the system tray , open a new window or adds a tab to the previous one, it hides and you keep working.

gfrenoy:
Thanks a lot, I made some tests in my environment but hit some issues we may want to discuss before publishing this change.

For sure I should have read the manual but to start with, I simply changed layout from "Traditional" to "Compact" and restarted Ásbrú.

I ended up in a situation where no tab is shown and no way to open the preference window :-(
(no tray icon active, ```asbru-cm --preference``` not working)

1)
I did not find the right-click thing on my own, not sure this is very intuitive...Can we have the preference button available in the "compact" layout as well ?  
To have more available space, we can maybe have only icons (also for clusters, scripts and PCC) ?

2)
If we edit ```pac.yml``` and remove the "layout: Compact" line, I would have expected that it reverts to the Traditional mode but it stayed in Compact mode.

3)
When I finally found a way to revert back to Traditional (by re-activating the icon tray in a previous version).  It has unchecked the "Put Terminal TABS in Main Window" so I did not revert back to my previous setting.

Can we avoid changing any other setting than the layout ?

4)
I could not re-enable "Put Terminal TABS in Main Window" in Compact mode ; every time I check that box and restart Ásbrú, if remains unchecked after the restart...

Is this behavior expected ?

hanspr:
Hi @gfrenoy 

>I ended up in a situation where no tab is shown and no way to open the preference window :-(
>(no tray icon active, asbru-cm --preference not working)

I assume that you are testing in some version of ubuntu with gnome-shell?, or none is a no gnome-shell environment?

I need to know the combination of your original settings to end up without tray and window, that would happen if you are on gnome-shell and the minimize to tray, start minimized were active or some other that I'm not taking into consideration. Could you share your general options for main and terminal and confirm if it is gnome-shell or not. So I can reproduce and see which combinations avoid

I'm very confused with gnome-shell because I do not see a tray icon available, how do you interact with asbru under gnome-shell? Or do you have a tray icon, how do you manage to have one if the PACTrayunity.pm is broken?

> I did not find the right-click thing on my own, not sure this is very intuitive...

Guess because you did not have any active window or active tray.

> Can we have the preference button available in the "compact" layout as well ?

I have done that

> To have more available space, we can maybe have only icons (also for clusters, scripts and PCC) ?
> 

I have done that

> If we edit pac.yml and remove the "layout: Compact" line, I would have expected that it reverts to the Traditional mode but it stayed in Compact mode.
> 

That is because you had to Remove (or Edit and change to Traditional) and remove pack.nfreeze. Asbru gives preference to this file if exists and loads the previous configuration from there.

> When I finally found a way to revert back to Traditional (by re-activating the icon tray in a previous version). It has unchecked the "Put Terminal TABS in Main Window" so I did not revert back to my previous setting.
> 

I'm working on saving the original configuration. I'm testing.

> Can we avoid changing any other setting than the layout ?
> 

Nope, because that is what it lead to your error. If the user is under gnome shell and forgets to inactivate the incompatible options, will create issues like:

* No available main window, what you experience. And I need to know which setting caused this to happen to you
* The windows would appear on the hidden main window (case of place tabs in main window) and you would not see it.

If you are move familiar with gnome-shell because you work on it daily, it would be a lot of help if you could check the code on the layout subroutine to adjust the parameters that work better it gnome-shell

> I could not re-enable "Put Terminal TABS in Main Window" in Compact mode ; every time I check that box and restart Ásbrú, if remains unchecked after the restart...

Is incompatible as explained in previous point. You can not have both

If you want a Compact mode then , you open windows or the tabs on a window, because the main window has been shrunk leaving no space to stack terminals in it.

But if you want tabs in the main window you should stay in the Traditional layout option.

----

I'll inform you when I finish all the changes so you can do a second test.

hanspr:
I just released the round 2, have implemented and tested changes requested.

But I was not able to reproduce in gnome shell that the window and tray icons do not show.  I need more information from your settings and environment to be able to reproduce and see how to handle that situation.

gfrenoy:
Thanks Hans :)  It works better now : I can see the terminals and I believe it works as expected.

I'm still concerned about changing the value of some options without the explicit understanding of the user.  See below.

Let me try to address the questions, if I forgot one, let me know:

> Gnome-shell

I'm running Ubuntu 18.04 with Gnome-Shell installed, indeed.

> Tray icon

I'm not 100% sure but maybe that option was simply not activated before I switched to 'Compact'.

When tray icon is activated, I see it because I have the [Unite extension](https://extensions.gnome.org/extension/1287/unite/) installed that will move the legacy icons to the top panel.

>> "Put Terminal TABS in Main Window" in Compact mode 
> Is incompatible as explained in previous point. You can not have both

Can't we just ignore "Put Terminal TABS in Main Window" when 'Compact' mode is enabled ; instead of changing the value ?

And prevent any change to it when 'Compact' mode is enabled ?

hanspr:
The problem is that "Put Terminal TABS in Main Window" is tested on many events, some of them on drag and drop operations, if I change the setting I'm 100% sure that it will work because it will react correctly to any event (known or unknown to me).

On the other hand, I have followed the code and it could hit with a new issue in points that I do not expect.

The settings are saved, and reverted back, I have tested so the user will never know. And I feel that in the worst case scenario, he will have to reactivate again.

From your description, the most challenging point is know if there is a tray icon available. If there is, I should not disable tray icon, if the user has no tray icon capability, then I should disable it to avoid the lost window.

Could you check : `ps xa` to see if there is a running process that will allow me to find if the tray option is available so I can handle this situation better?

hanspr:
I'm trying to install the Unite extension, to be able to test with it

hanspr:
The only way I find how to detect the tray is indirectly by testing the existence of the extension in the extensions folder, I will try with that and not disable the tray when I see the extension installed

gfrenoy:
> Could you check : ps xa to see if there is a running process that will allow me to find if the tray option is available so I can handle this situation better?

Sorry you lost me ... I'm not but what you mean with "the tray option is available"...?

In my current understanding : either we are running with gnome-shell and icons should be shown using a particular way ; either we are using the legacy gnome and icons are shown using another way.

Today Ásbrú only supports the old way so if you want to see the tray ; you'll to use an extension like Unite.

So if I got it correctly, we just need to detect if we are using gnome-shell or not and create the tray icon accordingly.  But that should be done in another issue ... For our layout issue here, it should have no impact, should it ?

hanspr:
The issue is:

I can not assume that all users are installing the Unite extension, to be able to see the tray icon.

For example, I'm not familiar with ubuntu and gnome-shell, so after I installed asbru, the first thing I noticed is that the window was lost and the icon tray was not accessible.

Just today, with your preivous comment, I learned that there is an extension that allows the use of the tray old system tray icon.

So, if there are more people like me : unfamiliar with gnome-shell, and unfamiliar with the existence of the Unite extension, I have to make sure that this particular users do not end up without access to asbru.

And at the same time, correctly handle users like you, that are familiar with gnome-shell and have installed the Unite extension, to still use it without interference. Or if the user eventually learns the existence of the Unite extension, and installs it later, that will work as expected after installing.

I finishing this changes, hope I have explained my self better.

B1tl0rd:
There are several extensions that have similar functions for legay tray, but some stop working, more than anything else for the gnome shell changes.

https://bugzilla.gnome.org/show_bug.cgi?id=785956

https://github.com/phocean/TopIcons-plus/issues/91

https://github.com/phocean/TopIcons-plus


I didn't know the "extension Unite" ... right now I can't install virtual machines. my default environment is kde / plasma

hanspr:
@gfrenoy 

This is round 3, and I think I managed to handle the events previously described, and have control over the tray icon in gnome-shell.

There is a small bug in my ubuntu VM, but sometimes the asbru tray icon does not show, the space is there and interaction is available, but the icon does not show.

imagen

And if I quit and open again, it shows again or may not show, but this does not have to do with my implementation, I guess all desktop environments, tray icons sometimes get lost. It happens to me in cinnamon some times with some icons.

imagen

gfrenoy:
> I can not assume that all users are installing the Unite extension, to be able to see the tray icon.

Indeed.  And as @B1tl0rd  mentioned, there are several of them.

So that's what I described as "the icon has to be shown using a particular way"... But I've no clue on how to do this :-/

>  I have to make sure that this particular users do not end up without access to asbru.

Well, the tray icon is really not necessary to use Ásbrú.  Personally, I don't use it much.  I guess you assume you need it because you minimize the tree after you open a connection.

What is an issue is that if you chose to have it ; it is not displayed !  I see other applications on my computer (like [psensor](http://wpitchoune.net/psensor/) or [KeePassXC](https://keepassxc.org/)) for which the tray icon is displayed, even without Unite.  So there must be a way :-/

> There is a small bug in my ubuntu VM, but sometimes the asbru tray icon does not show

Yes I do confirm.  I never tried to troubleshoot this further and this is the "legacy" way so I thought it was better to investigate the new way...

hanspr:
Hi @gfrenoy 

How is implemented right now in gnome-shell is:

* If extension Unite exists, then it will behave like any other desktop. It let the desktop to control the icon and honour your tray settings in the configuration file.
* If the extension Unite is no installed. The interaction will be to "hide" or "unhide" the window to the system tray, so you do not loose access to asbru, by having a window that disappears and no way to recover it.

I think if you test with both cases (Unite installed and not installed) you should be able to interact with the Compact mode in a normal way in both cases.

hanspr:
Do you want me to remove this PR, and reimplement again with utf8 changes?

gfrenoy:
No that's fine, I merged the recent changes so it should be up-to-date.

One thing I noticed when in minimal mode : we should probably remove the "info" and "Show/Hide bar" buttons ?

image

The history button did not work well either but that is not related to the minimal mode, I think.

hanspr:
I did not want to push my view of a compact window too much into users. In this particular case, I do not know if users like or not the status bar. I disable the status bar completely, I do not see any actual usefulness in it.

You can disable 'show statusbar'

imagen

But if you think it would be a good idea to inforce it into users, just add the line

`$$self{_CFG}{'defaults'}{'terminal show status bar'} = 0;`

Together with the other ones in PACMain.pm

```
if ($layout eq 'Compact') {
    $$self{_CFG}{'defaults'}{'tabs in main window'} = 0;
    $$self{_CFG}{'defaults'}{'auto save'} = 1;
    $$self{_CFG}{'defaults'}{'terminal show status bar'} = 0;
```

This is the boundary I did not want to cross. I just forced the ones that I know would create operational conflicts, not cosmetic ones.

We could document in the wiki the best way to take advantage of a compact mode and the recommended settings for tabbed and window oriented users. And let the users decide what they want.

gfrenoy:
The main goal is to bring new features without confusion.  So we should reduce the number of options and the number of "clicks" that a required to make it working.

Is there a way to "mask" options without changing them ?

I mean: even if ```terminal show status bar``` is checked, can we maybe disable or hide that option when the "Compact" mode is selected and whatever the option is saved in the option, we set it to 0 internally (but without changing the settings) ?

Same for the other options...

hanspr:
As I mentioned before, some options are very simple and can be overridden, because they appear very little in the code.

Some have a larger range of interactions across many files. Those, it does not matter what technique you want to use, involves implementing it in all the points where is currently impacting decisions, validations, etc. That to me, increases the possibility to create undesired side effects.

I'm only moving the ones that have a large impact in usability and/or  have a large impact on the code.

* tabs in main window = off
* auto save = on
* In the particular case of gnome-shell
    * Disable or enable hiding to tray if there is no system tray icon

All other changes are implemented with conditional statements and hiding elements.

In this particular case of the status bar and its Icons, they only have impact on Terminal.pm, so I do not see any problem to not use the option directly.

In this case you can apply the other technique that I already used in PACMain.pm

At the end of sub _initGUI or start. Go through a list of element you do not want to be shown and hide them, so the user can not see them or interact with them. No options changed.

```
foreach my $e ('','') {
    $$self{_GUI}{$e}->hide();
}
```

hanspr:
Sorry I'm slow to this git thingy, I finally realized that you were working on this merge pushing changes.

I added some changes to handle dimensions better.

But must importantly I think I begin to get some of your points. I added the functionality that changes in the layout option, hide or show conflicting options so the user does not see them and activates them, thou in the code they still will be set to the safe value.

The options will hide depending on the combination of desktop environment and layout selected, as seen in next images

Compact hides conflictive options in Behaviour

imagen

and Tabs placement in look and feel

imagen

Traditional, reverts to have them available again

imagen

imagen

gfrenoy:
> Sorry I'm slow to this git thingy,

Please don't be so humble !  I wished everybody would be "slow" as you are :)  No, really, you're doing a wonderful job and helping us a lot :)

I'm running this new version now.  It looks really good and it's stable enough to work with.
I like the fact that you're hiding the options that are not relevant for the mode we are running with.  Great improvement :-)

Let me use it for a few days and I'll let you if I have more comments.  Hopefully others can also give it a try ...

Thanks again and read you soon !

hanspr:
Hi @B1tl0rd 

Could you narrow down when the messages show?, when you click on any specific button, click or right click on where, opening a session, entering or exiting preferences or clusters.

B1tl0rd:
when I open it starts by minimized default, when I click, it starts to show the following error
`Argument "button" isn't numeric in subroutine entry at /usr/lib/perl5/5.30/vendor_perl/Glib/Object/Introspection.pm line 67.
`
Then when I start looking at the host list and click to open one, I get this error
`(asbru-cm:206887): Gtk-WARNING **: 13:11:02.372: ../gtk/gtk/gtkrender.c:1032: invalid icon size '0'
(asbru-cm:206887): Gtk-WARNING **: 13:11:02.372: Failed to render icon
(asbru-cm:206887): Gtk-WARNING **: 13:11:02.372: Invalid icon size 0
`

hanspr:
I guess the pac.yml file has something to do with it, because I do not see any error like that. But unfortunately those errors do not help to find the lines involved in the problem.

I assume this version you are running is installed in some directory under your home, and not in /opt/asbru-cm

I have been ruminating the idea to be able to export the pac.yml file without credentials and personal data to be able to do tests when this situations arise, because is very difficult to replicate.

I have created this script, if you could copy your pac.yml to your home directory, run this script in the same directory, should leave a pac_debug.yml file with all confidential information removed.

[clean_pacyml.pl.zip](https://github.com/asbru-cm/asbru-cm/files/3847743/clean_pacyml.pl.zip)

Check if it works correctly, and if it does, upload the pac_debug.yml file to be able to analyse it and check if I can be used to reproduce the problem.

And I will find a way to create a button or something to eventually be able to export all the necessary data to debug issues with more accurate information from the setup of a user.

B1tl0rd:
@hanspr when importing it, when I open it asks me for a password.

hanspr:
I did not consider importing it just be able to see the structure, let me try and create a new version that does not generate that behaviour

hanspr:
Here is e new version that has some fixes, but I could not reproduce the password required.

[clean_pacyml.pl.zip](https://github.com/asbru-cm/asbru-cm/files/3848106/clean_pacyml.pl.zip)

I assume that your pac.yml is encripted, that you use the protection option?

If so, then I guess that instead of copy pac.yml, and run clean_pacyml.pl

You will have to use the "Export" option of Asbru, without any encription and then run clean_pacyml.pl over a non encrypted pac.yml file, so it can be cleaned.

What I did, was to clean my pac.yml and replace the pac.yml from a clean install for this one renamed to pac.yml, removed pac.nfreeze. And I get this

imagen

All titles renamed, user, passwords, hosts removed, but I can see all the other configurations applied.

hanspr:
@B1tl0rd   

Ok, This should work

[clean_pacyml.pl.zip](https://github.com/asbru-cm/asbru-cm/files/3848186/clean_pacyml.pl.zip)

I just realized that asbru does not encrypts any file, so it must be some grabled text that I left on the cleaning process.

Tell me if this version works

hanspr:
@B1tl0rd 

Sorry, I if you are using "use gui password" option, it will prompt for the password and because I remove it It will not work. So I have to set that option to false at the moment to clean up to be able to import.

Here is a new version of the clean up script.

[clean_pacyml.pl.zip](https://github.com/asbru-cm/asbru-cm/files/3848245/clean_pacyml.pl.zip)

Thanks for helping me, I think if we make it work I could use this script as a model to export data from users to get a better environment setup when working over bugs.

B1tl0rd:
okay. Here I pass the file with the last scripts you sent
[pac_debug.zip](https://github.com/asbru-cm/asbru-cm/files/3848317/pac_debug.zip)

hanspr:
@B1tl0rd ,

It worked!, I can reproduce the problem.

imagen

And your configuration looks like this

imagen

hanspr:
@B1tl0rd 

I located the source of the bug, is this setting, that is not working

imagen

If you uncheck show tooltips, the messages do not show any more and begins working, normally

Now, I have to find the bug.

hanspr:
@B1tl0rd 

It should be fixed now, changed:

* removed depreciated method in gtk3 (set icon from stock)
* added validation of existing path, when empty is hovering over empty rows
* added markup fix to allow for characters : <> ' " to be included in names
* tested that works with utf8 names

hanspr:
> I added some changes to handle dimensions better.
> 
> The options will hide depending on the combination of desktop environment and layout selected, as seen in next images
> 

Moved **Main Window Layout** option to the top, because it will hide / show available options after it changes.

I think it makes more sense to hide / show them before the user goes over those options.

imagen

B1tl0rd:
@hanspr 
Thanks friend, but I still can't open terminals. When I click to open a terminal without any error, this error gives it to what I interact with movements of the compact window.
image
image

hanspr:
Have not been able to replicate this time, but I made some adjustments tell me if this works. The issue seem to be with screenshots. 

If the error still there, try to show me the actions you are taking, using some screenshots so I can have a better idea how to reproduce

B1tl0rd:
image
image
image
Now it shows no error. when opening any host, but it does nothing. If I change to traditional mode, if it opens, but with single windows not joined, if I frame and uncheck terminals in the main window and hide list. the traditional way works well again.

B1tl0rd:
I created a new host, to rule out some problem from my node list, in compact mode. when I try to open I get this window.
already in the configuration of the hosts, he had placed everything, address, port, user and password.
image

hanspr:
I have managed to replicate the first problem.

That one is a different issue related to expect.

hanspr:
Hi @B1tl0rd 

The first issue should be fixed now. Changing from compact to traditional and back should work normally.

Test that one to confirm the terminals show correctly now.

B1tl0rd:
Excellent.
Now it works well. :-)
thanks

hanspr:
Now for the second problem check if your password is the one you typed or if it changed, edit the terminal configuration and check what was saved, maybe is a problem with the encondings change to utf8

B1tl0rd:
If it is already fixed.

Issue #266, opened 2019-11-08
Find option shows lots of weird characters


ferent:
**Describe the bug**
In a shell, right click and select "Find... ", Contents of current log shows quite a lot of not readable characters, that makes the ability to find quite difficult.

**To Reproduce**
1. Go to any open session
2. Right click and select "Find... "
3. See the weird characters

**Expected behavior**
The same content I see in the shell output

**Screenshots**
https://ibb.co/BBLd778

**Desktop (please complete the following information):**
 - OS: Ubuntu 
 - Version 18.04

gfrenoy:
Thanks for reporting the issue @ferent and thanks to @hanspr for the fix !

That "Find..." option looks way better now :)

ferent:
Playing a bit with the find improvement, it looks really good now, but there is an small bug:
when hitting the search for second time (either searching for the same string, or a different one) the second time always return 0 results. Next click works ok, but the following doesn't.

PR #267, opened 2019-11-08
Fix266

hanspr:
Fix to issue #266


gfrenoy:
To avoid mixing different issues, I cherry picked the fix from this pull and also apply the same "trick" for the find results (in the left part of the tree).

Thanks a lot, it really looks nice now :)

Issue #265, opened 2019-11-07
asbru crashed, config file empty


The-Incubus:
**After a crash the config file is empty**
Asbru crashed and when i tried to restart it, it didn't.

**To Reproduce**
After opening and closing many connections, the interfaced becomes grey and unresponsive; close the window (don't just kill it). Restarting Asbru fails; the error shown in console is:
```
Ásbrú started (asbru-cm) with PID 21371
WARNING: Could not load config file '/home/MY_USER/.config/pac/pac.yml': 
INFO: Used config file '/home/MY_USER/.config/pac/pac.dumper'
not a reference at /opt/asbru/lib/PACMain.pm line 2974.
```
The config file is empty:
```
$ ls -l /home/MY_USER/.config/pac/pac.yml /home/MY_USER/.config/pac/pac.dumper
-rw-r--r-- 1 MY_USER MY_GROUP 0 Nov  6 17:08 /home/MY_USER/.config/pac/pac.dumper
-rw-r--r-- 1 MY_USER MY_GROUP 0 Nov  6 17:08 /home/MY_USER/.config/pac/pac.yml

```

**Expected behavior**
Config file not empty.
Asbru working.

**Desktop**
 - OS: Debian GNU/Linux 10 (buster)
 - Version: 5.2.1

hanspr:
Hi @The-Incubus 

* Asbru has way too many options. Please provide your settings options. Problem can happen with one combination of settings but not others : Tabs on external window, Tabs on main window, stand alone windows, clusters, autosave on or off, save on exit on or off, etc. Please share the images of your settings. Is best if you can take images of the settings so no one is omitted on my settings to try to reproduce.
* If you are able to reproduce the grey unresponsive window, take an image and try to describe the steps as precise as possible.
* For example:
   * How many is many is many connections, 2,4. I need to know so I don't stop at six when your exercise was at 10 or 20.
    * How do you interact:
    * open, open, open, open, open, close, open, close, close , close, close
    * open,close, open,close, open,close, open,close,open,close, open,close...
    * open, open, open, open, work working, for 10, 20 minutes, close, close, close
* What Asbru version are you using gtk3 branch or the main one.

If you are on the main branch try to move to the gtk3 branch, is very stable to my liking and if the problem is still there it would be fixed on this branch now. And then back ported if possible.

You will find the gtk3 branch and instructions to install in:

https://github.com/asbru-cm/asbru-cm/tree/gtk3

[...]

PR #270, opened 2019-11-08
Partial Fix265

hanspr:
This is a partial fix to issue #265 

While this does not fix the problem of the greyed asbru window.

It does provide a safety mechanism to recover a lost configuration file as requested by the user.

This safety mechanism will help to recover a known good configuration in case of any unknown bugs that could cause the lost of the main configuration file.


gfrenoy:
Maybe it's a bit too much for what you initially had in mind but to avoid any automatic action that would make things even worse, I would have seen this as follow:

1) If (and only if) the configuration file is properly loaded, we make a backup copy

2) By default, the backup location is a subdirectory of the Ásbrú configuration directory but one can configure another location (on another physical disk / network drive).  We can also configure the number of backup files we want to keep.

3) If we cannot load the configuration file, we show a message to the user asking what to do:
  a) Recover the last backup copy (if it exists)
  b) Do nothing (and let the user manage this)
  c) Start from an empty configuration

For a) and c), before deleting the configuration, we also do a copy before erasing the corrupted file.

hanspr:
Added 10 backups

* Test configuration file is readable and can be loaded.
* If errors, start searching for previous backups to recover the most recent one.
* Any damaged file is renamed to ".bad". So nothing is removed.
* Save into different directory, I'll have to leave it to you or for a second version later.
* Open a window asking for actions, will leave it to you.
    * Would be easier to notify after asbru is loaded saying that a recovery was done, that the user may check files .bad in the directory if he wants to take any manual actions to recover any old data, etc.

I run tests by editing and ruin the main file and several backup files, the application recovers from the must recent one that has no errors transparently.

gfrenoy:
Thanks a lot, I worked a bit more on this.

- Save files in a specific ```bak``` directory
- Do the check after directory creation
- Use a timestamp when keep a "bad" configuration
- First roll old files then create the .0 copy
- Skip check if there is no configuration file (typically when you first start Ásbrú)
- Only move to ad "bad" file if another configuration has been found (other left it untouched)
- Do not include ```getDesktopEnvironment``` (this is not related)
- Some cosmetic changes

Is there anybody to make more tests on this before I merge it back to the main gtk3 branch ?

PR #278, opened 2019-11-16
Proposal Export debug information

hanspr:
Added additional option to Export Settings to allow exporting debugin information.

imagen

This will allow to import users data without personal information and help debugging when the problem is related to users settings. Many times I have seen that some combination of settings trigger bugs that we are not aware of.

When we import the settings it will look like this

image

This has help me recently to find some bugs that were related to setting configurations.

In addition at the end of the file I add environment variables that would help to know better the user environment.

This is a first proposal, I would like that people tests that the export process really removes personal data and that no confidential information is published by mistake.

So we add more cleaning code if necessary. 

The file is a plain text file so the user can check the information collected and remove any that he/she does not wants to share

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 *.sw?
 *.*~
+lib/lib
+lib/res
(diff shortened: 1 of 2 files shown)


gfrenoy:
Interesting !  Thanks.

I was just thinking about the best label for this.  Maybe "Anonymized data" is clearer on this objective of this export format ?

Should we also add a disclaimer ?  Saying that even we did care very much about removing any personal data, everyone needs to review the content before sending it to anyone else and we do not engage our responsibility...?

And about the "YAML" and "Perl" Data, I always found this confusing ... Do you know why we have both ?  And do we really need that ?

hanspr:
I'm sure the perl data option has its value (David did a lot of useful stuff) so I commented in case we learn it later.

Added the disclaimer and option to abort, edit to your liking

imagen

PR #279, opened 2019-11-16
Automatically create configuration backups

hanspr:
Sorry, I was cleaning my repo off old branches and ended up smacking this one.

I added create bak directory if does not exits


gfrenoy:
Ok.  It looks good.  So are we ready to merge this change ?

Had anyone else the opportunity to test it ?

@hanspr Do we really need to have ```lib/lib``` and ```lib/res``` ignored ?  Can't we delete them after the compilation and/or move them outside the source directory ?

And in any case, if we really need it, better to have this outside this "backup" change.  This would be very confusing to find where this change happened ; better to have a separate commit.

hanspr:
> @hanspr Do we really need to have lib/lib and lib/res ignored ? Can't we delete them after the compilation and/or move them outside the source directory ?

I haven't found any other way around. One of the annoying things about Perl is the way it locates modules and includes them during a compile of a single library. And the the way asbru was originally designed, it complicates things even more.

imagen

I like to have my editor to compile on each save so I immediately know when I made a mistake. And believe me, I tried many setups tricks, spend several hours. I ended up nesting lib inside lib (with main files inside lib/lib and res inside lib to be able to do so). An not only that, the nested libraries, I had to remove all use strict and warnings, to avoid more errors on the original. And I have an old extra copy of asbru in /home/user/perl/asbru and environmental `export PERL5LIB=/home/user/perl/asbru`, to actually do  everything work.

So I do not know any other way around it, really.

But is not the porpoise of .gitgnore to help do this kind of stuff?

If it really complicates things that much, I will have to create a script to move stuff out and in on every push.

How is your setup?

gfrenoy:
> How is your setup?

I don't see those files and actually I don't really understand what you mean with "compiling" Perl code :-/  It looks like it is not compiled to me.... But anyway, I made a separate commit for the ```.gitignore``` file, that's OK.

hanspr:
My bad, "perl check", `perl -c file.pl`

My editor runs a perl check on the file after each save. So I can catch any syntax errors I made.

PR #284, opened 2019-11-24
Fix179

hanspr:
Added coding standard, encoding utf8

Added next sequence local port to avoid duplicates.


gfrenoy:
Thanks again for your help Hans :)

And if you could separate the commits for the code cleaning and the bug fix, that would really help understanding what is happening...

PR #286, opened 2019-11-29
Add layouts

hanspr:
Made a mistake again, deleted branch by mistake.

Proposal in reference to issue #262


hanspr:
@gfrenoy 

I just merged all the last changes, I been using the option all the time, have not found any issues.

Do you think is ready, or we should change something?

gfrenoy:
Indeed, waited long enough ! Let's go :-)

Issue #232, opened 2019-10-19
<-= DISCONNECTED (PRESS <ENTER> TO RECONNECT)  issue in gtk3


christospelekis:
Hello,
i have install xubuntu 19.10 in my laptop and then i have install asbru gtk3.
i have import some connections (ssh) from the export i have do before.
the most of the times, when i connect with ssh to a server i have receive the error:
<-= DISCONNECTED (PRESS <ENTER> TO RECONNECT) (Sat Oct 19 22:19:40 2019)
i click in enter to reconnect and after the connection i receive immediately the same error. And again and again and again. If i close asbru then i can connect for a while but after some time the problem happen again.
I have no issues with my network. Also i can connect find with the console without problems.
  
Is any log i can provide in order to find the issue?

gfrenoy:
1) Start ```./asbru-cm``` from the commande line and look at any message there
2) Activate the ```Expect DEBUG``` option in Preferences > Terminal Options > Advanced
3) Run the command outside Ásbrú (see in Connection > Advanced Parameters > Get Command Line)

christospelekis:
hi,
here is the results:
1. no anything related to console, just the:
INFO: Ásbrú started (asbru-cm) with PID 4487
INFO: Used config file '/home/user/.config/pac/pac.nfreeze'
INFO: Using Gnome tray icon

2. i have activate the Expect DEBUG but i just have see the same message during disconnection, without anything more.

3. Outside of asbru (using the same command, ssh -p 726 -2 -4 -x -C -l root X.X.X.X ) the connection work fine.
Thank you

gfrenoy:
Can you share what you see with ```2``` ?  That would help understanding what is happening for you.

From what you describe, I would suspected that the ```VTE``` terminal is not created correctly but that does not match what you first describe.  I mean, in that case, it would not be "most of the times" but "always"...

Can you confirm it sometimes work (or not) ?  Can you compare the output when it works and when it does not work ?

christospelekis:
hi,
here is a complete debug log. working for a while, but in the end you can see the disconnect.

[pac_conn(6690)][host1.example.com][host1.example.com - datacenter1]: !!_PAC_AUTH_[pac_PID{6499}_n3]!!
[pac_conn(6690)][host1.example.com][host1.example.com - datacenter1]: SPAWNING:ssh -p 2200   -x -l root host1.example.com
[...]

PR #287, opened 2019-11-29
Fix232

hanspr:
Fix issue #232 

Added delay to winch events, in particular xubuntu version 19.2, has some issues and the response is slower than normal.

Please test this branch before fully accept, have not seen any side effects.


hanspr:
Added definitive solution provided by [Perl Monks](https://www.perlmonks.org/?node_id=11109349)

PR #292, opened 2019-12-01
Fix layout

hanspr:
Fixes problem of group names shown incorrectly on tray right click.

Guess there was some merge error because the groups name was created 3 times


hanspr:
Removed the forced start minimized in compact mode.

hanspr:
Removed some old code from utf8icons caused bold markup not to work on group names

Issue #295, opened 2019-12-01
Copy password from another connection defined in the Connections List


popxunga:
Hi ...
Suppose that you have an SSH session established to host A.
Then you need to SCP a file from host B into host A and that authentication with password is required.
It would be nice to have an option in the context menu of the connection B to easily copy its password into the clipboard and then simply paste it while doing the SCP authentication on host A. 
Currently to achieve the same, we have to open the connection definitions of host B, expose the password field, select and copy the password, close the connection definitions window and finally we can paste the password while we execute the SCP command in session A.
Thanks.

gfrenoy:
As passwords are not properly encrypted in Ásbrú, I'm not sure we should encourage its usage ... Maybe working on a better integration with KeePassXC (see #65) would be more meaningful ?

popxunga:
I think that passwords are still widely used in most situations.
A better integration with KeepassXC is needed to. I'll l use for sure when it becomes available but until that happens, having what I'm asking for would be nice if the development effort to achieve it is not too much.

hanspr:
This should be closed

gfrenoy:
I tried this and actually, I would expect to have the "Copy connection password" on a right-click from the connection tree as well, don't you think ?

Said differently : I don't want to open a terminal or open the connection details to get the password.

gfrenoy:
Nevermind, it's already there...

gfrenoy:
@popxunga let us know if the latest snapshot is meeting your expectations.

popxunga:
Hi @gfrenoy  ... Just tried it ... and yes it does ! Thanks a lot :-)

PR #298, opened 2019-12-02
added feature #295 copy password, passphrase from terminal, connections

hanspr:
Added option to right clic menus on : terminal and connections.

Connections

imagen

Terminal

imagen

### DIFF
--- a/lib/PACMain.pm
+++ b/lib/PACMain.pm
@@ -2717,10 +2717,7 @@ sub _treeConnections_menu {
             shortcut => '',
             sensitive => 1,
             code => sub {
-                my $clipboard = Gtk3::Clipboard::get(Gtk3::Gdk::Atom::intern('CLIPBOARD', 0));
-                my $clip = _copyPASS($sel[0]);
-                use bytes;
-                $clipboard->set_text($clip,length($clip));
+                _copyPASS($sel[0]);
             }
         });
     };
--- a/lib/PACMain.pm
+++ b/lib/PACMain.pm
@@ -2717,13 +2717,8 @@ sub _treeConnections_menu {
             shortcut => '',
             sensitive => 1,
             code => sub {
-                my $clip = '';
                 my $clipboard = Gtk3::Clipboard::get(Gtk3::Gdk::Atom::intern('CLIPBOARD', 0));
-                if ($$self{_CFG}{environments}{$sel[0]}{'passphrase'} ne '') {
-                    $clip = $$self{_CFG}{environments}{$sel[0]}{'passphrase'};
-                } else {
-                    $clip = $$self{_CFG}{environments}{$sel[0]}{'pass'};
-                }
+                my $clip = _copyPASS($sel[0]);
                 use bytes;
                 $clipboard->set_text($clip,length($clip));
             }
--- a/lib/PACMain.pm
+++ b/lib/PACMain.pm
@@ -2714,6 +2714,7 @@ sub _treeConnections_menu {
         push(@tree_menu_items, {
             label => 'Copy Password',
             stockicon => 'gtk-copy',
+            shortcut => '',
             sensitive => 1,
             code => sub {
                 my $clip = '';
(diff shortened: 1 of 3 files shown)


B1tl0rd:
would have to identify it with another name. so that it does not conflict with the option that comes by default. which is to paste the password of the current session
image

hanspr:
@B1tl0rd , are you sure you are testing this branch? 

My images above and the one I just took, show like Copy Connection Password, not paste.

imagen

Though in my version checks for password or passphrase because you use one or the other.

But the original paste was just checking for passwords, so now it checks for passphrase too. But in any case, should show one row for copy and another for paste, if it shows only paste I think you are not testing this branch.

B1tl0rd:
No, I have not tried this change yet, I had only detailed this image.

gfrenoy:
Hans, don't we want to "factorize" the code a bit ?

And do we really need it in the terminal ?  I would assume that the initial request was only to have it in the connection list ...

hanspr:
> don't we want to "factorize" the code a bit ?

I don't know if I understand correctly this expression, if it is related to the fact that I'm using one test to copy pass or passphrase. I did it because I think you can not have both at the same time, the terminal edition activates one or the other.

imagen

Of course, if the users start using the terminal configuration as a password manager then we would need to separate. But initially it makes no sense to me to have both and use Asbru as a Password manager. But I understand the convenience to copy directly from within the application.

If this is not what you meant, then please explain it to me.

> Available in terminal

Reduces a step having to go to the list, but it can be removed, I do not mind. Is just an extra convenience. But if you think is unsafe in case of an unattended computer, then we can remove it, or we could consult with the user that made the request.

gfrenoy:
>> don't we want to "factorize" the code a bit ?
> I don't know if I understand correctly this expression

Sorry about the [algebra jargon](https://en.wikipedia.org/wiki/Factorization) ; I meant that I see twice the same piece of code:
image

I was wondering if we can make one single function out of this ?

I know this has not been done a lot so far in the code we definitely need, if possible, to avoid copy/paste of code snippets.

> Available in terminal

If I get the use case correctly, you are in "terminal B" and wanting to have the password of "terminal A" so you'll need to get that password from another connection, anyway.

Having the option to paste the password of the *current* connection is likely another use case.

hanspr:
I will create the function to do so in a minute

Original text

>Suppose that you have an SSH session established to host A.
>Then you need to SCP a file from host B into host A and that authentication with password is required.
>It would be nice to have an option in the context menu of the connection B to easily copy its >password into the clipboard and then simply paste it while doing the SCP authentication on host A.

This request is very complicated to achieve, I understand it as:

* I already have terminal A open.
* Want to SCP into A from B to put a file or SCP from B to A to pull a file.
* Have a context menu in Terminal B (this is a terminal) to copy password from A. This is too cumbersome. Using B to access A. Way to many jumps inside the application to accomplish that and locate the one you want to pull from.

But to me there are 2 short paths to get the password you want.

From an existing opened terminal, or the list.

Use case, example that happens to me very often:

* I open an ssh connection with asbru
* Go to my file manager (outside asbru) and type my scp://.... to open a remote folder on my desktop file manger.

imagen

*  The terminal is already available and accessible to reach for the password, and pass it to the file manager

hanspr:
Code has been "factorized" :factory:

B1tl0rd:
I think he means, with scp it's by command line from the terminal example.
scp  file.tar.gz user@remotehost:/home/user/
then be able to paste the password of the remote host

hanspr:
@B1tl0rd , yep you right.

My image was wrong. I actually do : sftp://server/... and have to paste the passphrase from my already opened terminal.

Still think the convenience to have access to the password from the terminal or list, is for a wider use cases besides scp: rsync, sftp, to create a new certificate, to open a certificate locally, etc.

Of course I use KeePassXC, but the convenience of having another method is not bad after all.

Even if we connect Asbru to KeePassXC, we could pull it from there too. To avoid having to jump out of the application.

gfrenoy:
Thanks for the great work Hans :)

Just a quick note: Klaas has merged the gtk3 branch to master so, unless you want to start a "huge" change, you should now branch from the master to create your new PR.

hanspr:
@gfrenoy 

Got it. Good to know you have decided the new version was ready.

Issue #317, opened 2019-12-21
Some regressions in gtk3


merikz:
Due to #315 I took the liberty to review the code changes for the refactoring to remove "unless" from the code base, and found some suspicious looking code.
 For some of them I could find actual faulty behavior, but for others I haven't been able to trigger the suspicious code.

Anyway, I created a commit in my fork with comments, see commit https://github.com/merikz/asbru-cm/commit/b75c9e87d703394cd806c80be4b3bb044afcc9a0.
This is just to document what I found, it does not attempt to fix anything.

The commits that introduced the changes claims to apply coding standard #217, but maybe some of the logic changes actually correct faults? 
Maybe @hanspr can have a look?
I guess I can try to come up with a PR to correct the faulty behavior I've seen, if needed. 
What do you think?

hanspr:
Hi @merikz 

I guess I did have some slip ups, it was a large rewrite.

`unless scalar @x ==1" is not same as "if scalar @x ==0"`

guess the safer change is to have

`unless scalar @x ==1" to "if scalar @x!=1"`

The others were mistakes too, thank you for spot them.

I'm sending a PR in a second, and hope you can give a your opinion on the changes, thanks again.

gfrenoy:
Thanks to both of both for keeping the quality of the code to a high level :)

merikz:
It was a large rewrite, indeed, but needed.Thanks for the effort. Inverted negations does make you head explode.
I just comment on the PR.

PR #318, opened 2019-12-21
revised unless statements, caused regressions

hanspr:
Fixes spoted by @merikz on #317


merikz:
The regression at line 1248 is that the "IN CLUSTER" value in the GUI does not update anymore when a cluster with active terminals is deleted, and the user agreed to "Remove running terminals from deleted cluster" (a sentence which is a bit incomprehensible IMHO). But the terminals are are in fact no longer clustered.
Related: What happened to the "Running clusters" concept in Cluster Manager in the GTK3 re-write? Gone? By design choice or by accident? (There is still some code left for it.)

The (possible) regression at line 1280 seems to be some kind of protection against a race condition that I don't fully understand.

(Again, line numbers according to https://github.com/merikz/asbru-cm/commit/b75c9e87d703394cd806c80be4b3bb044afcc9a0)

hanspr:
Guess I missed it I'm travelling right now, will check later

Issue #327, opened 2020-01-03
Click on re-attached tab, open an empty window


nikuzz:
**Describe the bug**
When dragging a tab out and then right-click inside its terminal and click "Attach Window to main tab bar", any click on the same tab open an empty window.

**To Reproduce**
1. Open a "local" shell tab (ctrl+shift+t)
2. Detach this tab dragging it out of the main tab bar (or right click on tab, then "Detach TAB to a new Window")
3. Right-click inside its terminal and click "Attach Window to main tab bar"
4. Click on "info" tab
5. Click back on the "local" tab
6. An empty window will appear

**Expected behavior**
The empty window shouldn't appear.

**Screenshots**
![Captura de pantalla de 2020-01-03 12-08-44](https://user-images.githubusercontent.com/17493820/71720407-ef313780-2e21-11ea-9d75-afd587e08708.png)

**Desktop:**
 - OS: Ubuntu 19.10
 - Version v.6.0.0

gfrenoy:
Thanks for your feedback Nico.  I do confirm the buggy behavior... No quick fix right now, I'll need to investigate further.

gfrenoy:
Ok I think I found the issue !  It will be fixed with the next commit that will be available in a few minutes in the snapshot release (and will be available in the next release 6.0.1).

Thanks again for the report and your support to the project !

PR #334, opened 2020-01-04
fixed regression tabbed terminals

hanspr:
Fix regression tabbed terminals not showing

### DIFF
--- a/lib/PACTerminal.pm
+++ b/lib/PACTerminal.pm
@@ -829,6 +829,7 @@ sub _initGUI {
         _setupTabDND($self);
 
         $tabs->append_page($$self{_GUI}{_VBOX}, $$self{_GUI}{_TABLBL});
+        $tabs->show_all();
         $$self{_GUI}{_VBOX}->show_all();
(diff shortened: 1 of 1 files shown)


gfrenoy:
Reverting this will break #327 !

We should understand why you have that regression and find the right way to fix it for both cases.

Can you elaborate the case where SSH terminal does not work for you ?

gfrenoy:
Ok I understand the problem ... You are using terminals outside the main window !  Let me find the right "show_all" ; we should not do that for all terminals ...

hanspr:
Yes, so maybe with an if to test for that should make it work

hanspr:
I'll wait for you because I'm about to add another PR

gfrenoy:
Does "show()" instead of "show_all()" works for you ?

(it looks to fix both issues here...)

gfrenoy:
@hanspr Ok just pushed https://github.com/asbru-cm/asbru-cm/commit/31d4c55b281efa196e25187a2da7730232e8df8c ; looks to work fine for me, let me know if you should still encounter some issues with terminals outside the main window.

hanspr:
Yes it works, thanks

Issue #323, opened 2019-12-30
Remove Net::Proxy, replace with ssh option parameters


hanspr:
Hi @gfrenoy 

There are 3 bugs related to net::proxy: #218 , #67  and #280

net::proxy, does not work properly in the case of issue #280. I think its porpoise was to give a functionality that was not available through ssh some years back.

But now, with command options we can control the 2 case scenarios and remove dependency to net::proxy

* #67 : -J option
* #280: -o ProxyCommand='nc -x 127.0.0.1:5000 %h %p'

Would not be backward compatible with older ssh versions as noted on #67, but with a tunnel and ProxyCommand, you can implement the same solution in this cases, so the user could use the available option for his own ssh version.

The "Proxy" tab could be renamed to "Network settings", as in #67, and we could add the options to the connection strings in pac_conn.

The tab should have just 4 options

* No proxy
* Use global proxy
* Use proxy : host,port,user,pass
* Use Jump host
    * One jump host configuration as in image in #67, or
    * Allow configuration of multiple jumps? or a max of 3 jumps to make it easier to implement the interface?

What do you think?, should we remove net::proxy and implement everything with options?

KlaasT:
I think that could work. 

We have to remove Net::Proxy. As in #218 the dependency is hardly supported with newer Perl versions.

KlaasT:
Edit: Quickly edited your opener to link the issues.

gfrenoy:
Yes that'd be great if we can get rid of this dependency.  I tried to understand how it works but I failed so I suspect nobody is using it, anyway... Have a "cleaner" (less complicated) code base will definitely help !

KlaasT:
@gfrenoy I tried to undertand Net::Proxy in the past and had a look if it can be replaced. A replacement using Perl libraries won't be simple. Basically you would have to write an own library which we would have to support. Not the way I would like to go.

In case of older SSH versions not working we could add a check to disable these new options if OpenSSH is too old.

hanspr:
If there are no further comments or requests, we should close this issue and related ones: #218 , #67 and #280

gfrenoy:
For #67, see checklist here : https://github.com/asbru-cm/asbru-cm/issues/67#issuecomment-570896646

hanspr:
I think the list is complete we should close this.

gfrenoy:
[...]

Issue #280, opened 2019-11-19
SSH connection via SOCKS proxy not working


carlos-n:
**Describe the bug**
Once a SOCKS tunnel has been established, a new SSH connection using this tunnel doesn't work. 

**To Reproduce**
1. Define a new SSH connection against a server "A" configuring a "Dynamic Socks Proxy" in "SSH Options" section. We use "127.0.0.1" as "Bind Address" and "5000" as "Local Port".
After creating the connection, we establish the connection with server "A".
We can see our local server is listening in port 5000.

tcp    LISTEN   0        128                                                        127.0.0.1:5000       

We can open connections against servers using the SOCKS tunnel with Putty. 
So, the SOCKS tunnel is correctly created and working OK.

2. Define a new SSH connection to a server "B". In "Proxy" tab of "Connection" section we configure the option "Always use next proxy /HTTP/SOCKS) for this connection". We use "127.0.0.1" in "IP/Host" field and "5000" in "Port" field. We leave empty "User" and "Password" fields.
 
3. When trying to establish connection with server "B", we receive an error.

connect() failed with error 'Connection timed out'
ssh_exchange_identification: read: Connection reset by peer
pac_conn: Fatal IO error 22 (Invalid argument) on X server :0.

**Expected behavior**
Connection with server "B" should correctly established as it happens when you try it using Putty.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Linux Mint 19.02 64 bits
 - Version: Asbru v5.2.1

**Additional context**

gfrenoy:
I tried a few times to reproduce this but got stuck at step 2.  I just realized it's a bug in the gtk3 branch, see #291... I soon as I can get that working, I'll have another look.

In the meantime, did you check the SSH command line of the connection to server B ?  Does it work if you run it outside Ásbrú ?

gfrenoy:
I looked a bit more at this, here are some notes 

- Ásbú uses the perl module [Net::Proxy](https://metacpan.org/pod/Net::Proxy)
- When configuring Ásbrú as you did, it uses the following configuration of Net::Proxy to establish the proxy:

https://github.com/asbru-cm/asbru-cm/blob/1687e82bd340cdbf1296e8efec117d49f280b48a/lib/pac_conn#L848-L861

- Not sure to understand what it exactly do but, as a matter of fact, it does not connect to the proxy ...

[...]

Issue #218, opened 2019-10-14
libnet-proxy-perl dependency old and deprecated


figue:
I've been testing gtk3 build and checking dependencies saw that libnet-proxy-perl remains there but for me Net::Proxy seems old and deprecated upstream

https://metacpan.org/release/Net-Proxy

Can be replaced to somewhere else?

KlaasT:
Do you have any information where it states that the dependency is deprecated? I checked with the Debian repositories and even sid still seems to ship that package. Otherwise we would need to search for a replacement.

figue:
![image](https://user-images.githubusercontent.com/3238330/66750170-0d734100-ee8c-11e9-9e23-937430f9b837.png)
http://matrix.cpantesters.org/?dist=Net-Proxy+0.13

KlaasT:
As far as I can see it is not very easy to replace that library. I'm having a look on what could be done using LWP::UserAgent but I can't find a simple drop-in replacement for NET::Proxy.

figue:
In Archlinux I couldn't build Net::Proxy. And, as reported in matrix, probably doesn't work either. I didn't tried in Debian or other distributions, could you confirm that proxy as is works or is broken? Personally, I use custom attributes in freerdp to workaround.

KlaasT:
It is probably broken with modern Perl versions and if the dependency is missing.

gfrenoy:
Ásbrú does not have any dependency to Net::Proxy anymore.

Issue #67, opened 2018-03-19
Add support for the "jump host" functionality of SSH


Cris70:
Hi, it would be very nice to have a straightforward way to use the "jump host" (option -J) functionality of SSH.

I'm attaching a screenshot of the way it is implemented (GUI-wise) in MobaXterm, it may be an inspiration for how to model it in Asbrù.

![screenshot_20180319_120431](https://user-images.githubusercontent.com/5107224/37594175-c2978b54-2b74-11e8-8d41-50d541b23348.png)

gfrenoy:
Is there any requirement to be able to use that "-J" option ?  I mean: I don't see that option in my SSH (Ubuntu 16.04):
`ssh -v`
`    usage: ssh [-1246AaCfGgKkMNnqsTtVvXxYy] [-b bind_address] [-c cipher_spec]
           [-D [bind_address:]port] [-E log_file] [-e escape_char]
           [-F configfile] [-I pkcs11] [-i identity_file] [-L address]
           [-l login_name] [-m mac_spec] [-O ctl_cmd] [-o option] [-p port]
           [-Q query_option] [-R address] [-S ctl_path] [-W host:port]
           [-w local_tun[:remote_tun]] [user@]hostname [command]`

Cris70:
Which version of ssh?
I'm on openSUSE Tumbleweed here and I have it:

```
cris@PolariSuse [~]$ ssh -V
OpenSSH_7.6p1, OpenSSL 1.1.0g-fips  2 Nov 2017
```

```
cris@PolariSuse [~]$ ssh
usage: ssh [-46AaCfGgKkMNnqsTtVvXxYy] [-b bind_address] [-c cipher_spec]
           [-D [bind_address:]port] [-E log_file] [-e escape_char]
           [-F configfile] [-I pkcs11] [-i identity_file]
           [-J [user@]host[:port]] [-L address] [-l login_name] [-m mac_spec]
           [-O ctl_cmd] [-o option] [-p port] [-Q query_option] [-R address]
           [-S ctl_path] [-W host:port] [-w local_tun[:remote_tun]]
           [user@]hostname [command]
```

gfrenoy:
Looks like Bionic will be very welcome :)

    $ ssh -V
    OpenSSH_7.2p2 Ubuntu-4ubuntu2.4, OpenSSL 1.0.2g  1 Mar 2016

[...]

PR #335, opened 2020-01-04
Fix323 FOR TESTING

hanspr:
This is the first version to remove Net::Proxy from #323, and address #280, #67, #218 

This version has a print to show the connection string used to connect to test in case of errors, we need to remove after we have confirmation that the interface works properly.

__General Changes to the interface__

* General Settings
    * Renamed to networking
    * Changed logic to options
    * Added Jump Server global options
    * Renamed labels

imagen

* Terminal Connection Settings
    * Changed to networking
    * Added Jump Server
    * Renamed labels

imagen

__Currently Tested, case #280__

* Configure to first connection gateway machine, add Dynamic port

imagen

* Configure access to second machine connection, add connection information

imagen

* Setup networking matching Dynamic IP/PORT, setup on first terminal connection

imagen

__Test results__

Connection to intermediate machine, correct
imagen

Connection to machine in local network correct

imagen

---

__Case #67__

After a lot of reading, got to the conclusion that this method only works with passwordless connections. So to me the easiest setup would be to create entries in .ssh/config, similar to this

imagen

imagen

2 Entries should be created one for the Jump Server and the second for the target machine

And then configure Asbru like this

* Connection with to second machine used in .ssh/config

imagen

* Configure networking, with name of JumpSever used in .ssh/config

imagen

__Test results__

imagen

---

Please test, and confirm if it works for you.


gfrenoy:
Waw, great !  Our next milestone (https://github.com/asbru-cm/asbru-cm/milestone/2) is maybe closer than I initially thought :)  Thanks again for your great contributions :)

We should merge this the "[loki](https://github.com/asbru-cm/asbru-cm/tree/loki)" (unstable) branch so that we can keep the "master" ready for hot fixes.

Do you see how to do this and will you have time ?  Otherwise, I can have a look on my side.

One remark : the merge of the glade file will likely be tricky but you can just revert my recent changes, I will re-apply them afterwards.

hanspr:
I do not know how to change the merge to "loki", except to remove this PR and create it again, is there a simpler way? or should I remove it and send it again?

hanspr:
Found it

gfrenoy:
Ok, I merged your version and then re-apply my changes.

So we should have the "latest and greatest" version on loki now...

Let make some tests !

hanspr:
Ok, I think I have it updated, I'll wait for your feed back

gfrenoy:
Yes, you'll need to get the latest glade file.  It's actually a nightmare to follow up the changes ; it looks like glade is constantly changing things that I did not touch.

Anyay.

I started to look at the "jump" option but got confused by the "SSH config" ; shouldn't we just mimic what was initially suggested (#70).  I would assume that such a "jump host" would work with a SSH key ...

hanspr:
I think you refer to a different issue than #70 , think you refer to #67 

Well I wanted to be able to tell the difference between an .ssh/config entry and a user@server:port setup

Is different, you may rename it or change into 2 different lines, if you think it makes it clearer, but the concepts are different

This is related to this my reference

>After a lot of reading, got to the conclusion that this method only works with passwordless connections. So to me the easiest setup would be to create entries in .ssh/config, similar to this

SSH Config, tells me that I should create a command line like : ssh -J machine1 machine2

The other entries tell me to build : ssh -J user@machine1:port user@machine2:port

hanspr:
> would assume that such a "jump host" would work with a SSH key ...

Yes but again is not possible to pass that parameter with an -J option, so the option in #67, is useless.

You have to configure .ssh/config or create a rsa key, but can not build a command line to use it.

To build your rsa you should follow a similar approach as this: https://www.tecmint.com/ssh-passwordless-login-using-ssh-keygen-in-5-easy-steps/

But to me is a lot simpler to create .ssh/config entries, is by far more flexible.

hanspr:
For example, move the settings like this

imagen

So you can tell if you want to use an SSH config entry or a user@machine:port configuration.

This are some of the references I used to understand the -J option

https://www.tecmint.com/access-linux-server-using-a-jump-host/

Do you wan't me to move the glade file again? to make it look like the image?

gfrenoy:
I'm just thinking loud:

- Users of a GUI as Ásbrú probably don't want to know something about ssh_config ; they do expect the UI will help them making this as easy and straightforward as possible
- Can't we have only one "Host" field ?  If that host is defined in ssh_config, ssh will use it "automagically"
- In the tooltip, we can explain it can also be an IP address or a SSH config entry
- Port will need to remain empty (that should be the default, I would guess, not 8080)
- To use a SSH key, I indeed think we need to use the "legacy" ProxyCommand version (see all examples [there](https://en.wikibooks.org/wiki/OpenSSH/Cookbook/Proxies_and_Jump_Hosts#Passing_Through_One_or_More_Gateways_Using_ProxyJump))

hanspr:
Do you have time to chat, where, the link https://webchat.st-city.net/#asbru-cm is always down now

hanspr:
Think I have a very good solution, give me an hour

gfrenoy:
Yes but indeed, the St-City webchat seems to be down today ; I'll ask around if we can fix this.

In the meantime, you'll need to find another IRC client.  Just tried this one: https://kiwiirc.com/nextclient/ ; looks to work ... Server is irc.st-city.net ; channel #asbru-cm

PR #336, opened 2020-01-04
Netproxy remove

hanspr:
Round 2

Now you can configure both servers like any other connections, the only limitation is the issue with the keys must not have a password to work

Destination

imagen

Jump Server

imagen


hanspr:
Forgot to mention, what it does is it creates a ssh_config file, saves it in pac/tmp and opens the connection as:

`ssh -F path/to/config -J jumpserver server`

hanspr:
#323 

Added comment to relate to issue

Issue #210, opened 2019-10-11
[gtk3] VTE bold is bright option


egmontkob:
VTE 0.52 introduced a new API: `set_bold_is_bright()`, defaulting to true (the old behavior).

VTE 0.56 changed the default to false, meaning that the `\e[1m` sequence enables boldness only, but does not switch to brighter colors.

This was done in order to separate the concept of bold from the colors.

Some people might not like the new default, though, so you might want to consider adding a compatibility checkbox. (I'd truly appreciate if your default was the new, "false" state :))

Correspondingly:

PACTerminal.pm feeds red/green text to the terminal by printing `\e[1;31m` or `\e[1;32m`. These also become dark or bright depending on this setting.

You might want to make it immune to this setting. Bold+dark can be achieved using the 256-color notation `\e[1;38;5;1m` and `\e[1;38;5;2m` (this is somewhat specific to VTE, other emulators might also make it bright, but that doesn't matter here). Bold+bright can be achieved using `\e[1;91m` and `\e[1;92m`.

hanspr:
This should be closed

PR #337, opened 2020-01-05
Fix210

hanspr:
Fix to #210 
Added option in Preferences , Terminal Options , Look & Feel

imagen


gfrenoy:
That glade file is really a problem ... The merge was impossible to understand :-/  I assumed that PR was the last one that integrated the latest changes of fix233.  Let me know if I got it right...

hanspr:
Looks fine to me, lets hope not many errors come up later.

I do not know if there is a procedure we could follow to make future changes to the glade file less painful.

I use the application Glade to edit. But it looks to me that any minor change could create large changes in the resulting xml file like reordering nodes, added extra nodes.

Issue #343, opened 2020-01-06
Keepass does not work in latest


accessone67:
**Describe the bug**
After updating asbru to latest version It wont load if I use my keepass settings. I had to downgrade disable keepass and upgrade again, then asbru will load. But if I want to configure my keepass whit latest asbru I get error

**To Reproduce**
1. Go to asbru settings
2. Click on keepass and enableing it
3. See error
sJan  6 10:44:33 tamps asbru-cm.desktop[27882]: *** unhandled exception in callback:
Jan  6 10:44:33 tamps asbru-cm.desktop[27882]: ***   undefined value for mandatory argument 'text' encountered at /opt/asbru/lib/PACKeePass.pm line 279.
Jan  6 10:44:33 tamps asbru-cm.desktop[27882]: ***  ignoring at /usr/share/perl5/Gtk3.pm line 546.

If I dont disable keepass before upgrade I get this error after I want to startup asbru
Jan  6 10:41:12 tamps asbru-cm.desktop[27882]: *** unhandled exception in callback:
Jan  6 10:41:12 tamps asbru-cm.desktop[27882]: ***   undefined value for mandatory argument 'text' encountered at /opt/asbru/lib/PACKeePass.pm line 279.
Jan  6 10:41:12 tamps asbru-cm.desktop[27882]: ***  ignoring at /usr/share/perl5/Gtk3.pm line 546.

**Expected behavior**
To get keepass working like in old version

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Version: 18.04

gfrenoy:
I would presume it's 6.0.1 but can you confirm the version number you are using ?

accessone67:
I got this problem from bouth versions in 6. just did upgrade to 6.0.1-1 and erros is still there

Jan  7 09:25:37 tamps asbru-cm.desktop[1942]: INFO: Ásbrú started (asbru-cm) with PID 1942
Jan  7 09:25:37 tamps asbru-cm.desktop[1942]: INFO: Desktop environment detected : gnome-shell
Jan  7 09:25:38 tamps asbru-cm.desktop[1942]: INFO: Using config directory '/home/tambet.paide/.config/pac'
Jan  7 09:25:38 tamps asbru-cm.desktop[1942]: INFO: Used config file '/home/tambet.paide/.config/pac/pac.nfreeze'
[...]

Issue #65, opened 2018-03-08
Proper KeePass support


yennor:
Right now Asbru supports KeePass files, but the support is suboptimal.
- only old kdbx 3.1 file format is supported
- Asbru does not notice when passwords are changed or added to the keepass file.
- asbru insist on, that the password needs to be saved with the config files. That means it will be stored unencrypted. It would be better if asbru asks on start up for the keepass file password.

The best solution in general would propably be, if asbru could support the browser integration protocol of KeePassXc. It is a  little bit described here: 
https://keepassxc.org/docs/keepassxc-browser-migration/

Like that most above points are  solved. Additionally the user only needs to delock the database once.

what do you think?

gfrenoy:
I definitely support the idea :)  We need to link to issue #56 as well.

Now, how can we make progress on this ?  Is there anybody willing to work on this ?

Cris70:
Yes, please! Add the support for KeepassXC!! :+1:

Prototype-X:
The KeePassX password is stored in a configuration file and is easily decrypted.
Can you fix this?
1. Do not store the KeePassX password in the configuration
2. Ask for KeePassX password when starting asbru

Stunherald:
I second to that. Please add KeePassXC support :)

mbreskovec:
Hello guys, any progress in this case, or any time estimation? This is something what really makes me cry, when I need to change password for all my connections manually.

gfrenoy:
I'd love to see progress but I do fear the project will need your help ; anyone to follow-up on my last questions:

> how can we make progress on this ?
> Is there anybody willing to work on this ?

gfrenoy:
Oups.  Did not mean to close the issue ; sorry.

figue:
Recently I have been converting my kdb databases to version 4.0 and have suffered this issue, because only 3.1 is supported. Will be a very good addition to support new version as reported.

skipperTux:
With regards to [yennor's](https://github.com/asbru-cm/asbru-cm/issues/65#issue-303488703) comment I did some googling for KeePassXC integration via browser integration. This integration type solves a few of the above problems like

- no support of different file formats necessary
- no storing of Master-password necessary

because access of the password database is all going through the unlocked KeePassXC application using a proxy.

[...]

Issue #56, opened 2018-01-24
Add KeePassx key file support


Dodewaard:
Hello,

I would love to use the KeePassx feature, but our database uses an additional key file to unlock the database. 
Currently there seems to be no support to use a key file in addition to the master password to unlock the database. Maybe this could be a nice feature to add in the future.

Thanks for keeping this tool alive!

gfrenoy:
Ásbrú now supports KeePass key file ; testers are welcome on the [loki branch](https://github.com/asbru-cm/asbru-cm#testing-new-features).

PR #345, opened 2020-01-07
Fix65

hanspr:
## Support for KeePassXC #65 , #343 

The functionality was complete rewriten to address several requests related in the same 2 issues:

* Password no longer required in configuration
    * If password is not provided, will ask for password on load and retain in memory until asbru exits.
* User names and passwords are retrieved from database on the fly at the moment of connection
    * You can change your passwords in keepassxc and in the next connection the new one will be used
    * No passwords are retained in memory or files any more
    * User name and passwords are references now
* Previous implementations would extract absolutely all passwords even when they were not related to any connection, pulling out any other information that a user could have stored like : bank accounts, nips, etc.
    * The full password list was located in memory in a log submenu.
* Button in Connection edit, allows to search entries and doubleclick to select desired entry. The correct format will be inserted in the corresponding fields for user,password or user key and passphrase. Depending on the selected authentication option

imagen

__Search__

* You must type 3 characters or more to start a search

imagen

* During the configuration the application will detect : Version available and capabilities detected on that current version.

imagen

## Limitations found during integration:

* The original proposal was to use <field:uuid>, keepassxc does not allow to retreive the uuid, and last version of keepassxc-cli, no longer allow to search by uuid, so I had to drop it because is not reliable.
* Make sure all titles in keepassxc are different to be able to associate to a unique key


hanspr:
This should address #56 as well, not tested yet.

hanspr:
@gfrenoy 

Hi Gaëtan, I know we chat about to create a script to migrate current user configurations for keepassxc integration, but I checked the way the tags are created and is impossible, will be a very flawed migration.

The users will have to reassign the entries, but the effort will pay.

gfrenoy:
I don't use this feature so this is not an issue for me but for those who have a long list of entries ; that will be quite a pain !

How can we evaluate who is really using that feature ?  How many entries they have ?  How long would it take take to reconfigure each entry ?

To make it short: what would be the impact ?

hanspr:
Let me rethink the script, I said that I could not automate it (meaning fully automate it), but it does not mean that it could not be an interactive script.

Meaning that I could run over the file and look for the matched entries in the database, and if I find more than 2, then I ask the user to choose the correct key, similar to what the current implementation does in asbru, when it finds more than 1 matches it asks you to select the correct one. For that I would have to use the current KeePass.pm library.

Does it makes sense?

gfrenoy:
Yes, even if it's not fully automated, if there's a way to migrate that we document and explain, it means that the impact will be - hopefully - limited.

But in any case, it's not a change that we should "widely" push without making additional tests and prepare users' questions (in wiki pages or something).

I really think we need to finalize milestone 6.1 (https://github.com/asbru-cm/asbru-cm/milestone/2) and then see what we include in the next release.

PR #349, opened 2020-01-07
Fix65

hanspr:
## Support for KeePassXC #65 , #343 

The functionality was complete rewritten to address several requests related in the same 2 issues:

* Password no longer required in configuration file
    * If password is not provided, will ask for password on load and retain in memory until asbru exits.
* User names and passwords are retrieved from database on the fly at the moment of connection
    * You can change your passwords in keepassxc and in the next connection the new one will be used
    * No passwords are retained in memory or files any more
    * User name and passwords are references now
    * url can be used as a reference to get the host to connect
* Previous implementations would extract absolutely all passwords even when they were not related to any connection, pulling out any other information that a user could have stored like : bank accounts, nips, etc.
    * The full password list was located in memory in a log submenu. Difficult to follow.
    * Some entries could return more than one match and the user would be required to choose the correct one.
* Button in Connection edit, allows to search entries and double-click to select desired entry. The correct format will be inserted in the corresponding fields for user,password or user key and passphrase. Depending on the selected authentication option
* Right click on the hostname allows to select url from KeePassXC to use the reference in url as the hostname and retrieve at the time of connection.
* KeepassXC entries can be added to Exec, Expect and Pre and Post Exec. With right mouse.

imagen

__Search__

* You must type 3 characters or more to start a search

imagen

* During the configuration the application will detect : Version available and capabilities detected on that current version.

imagen

## Limitations found during integration:

* The original proposal was to use <field:uuid>, keepassxc does not allow to retrieve the uuid, and last version of keepassxc-cli, no longer allow to search by uuid, so I had to drop it because is not reliable.
* Make sure all titles in keepassxc are different when located in the same Path, to be able to associate to a unique key
* Interactive command was removed in the most recent version of the CLI, so it can no longer be considered as an option


hanspr:
Sorry my bad, wrong command executed

lib/pac_conn
```diff
@@ -1393,7 +1395,7 @@ elsif (! $MANUAL) {
                 ctrl("HOSTKEY:accepted by configuration (sent '$yes')");
                 exp_continue;
             } else {
-                my $val = wEnterValue(undef, $match, undef) // return undef;
+                my $val = wEnterValue(undef, $match, "Answere yes/no") // return undef;
```
merikz:
```suggestion
                my $val = wEnterValue(undef, $match, "Answer yes/no") // return undef;
```

hanspr: done

skipperTux:
With regards to [this comment](https://github.com/asbru-cm/asbru-cm/issues/65#issuecomment-571105956): Is there any target date when this PR is merged into loki, or shall we test it by cloning the repository?

hanspr:
I think it is a good idea to test. This breaks previous configurations. So copy your current configuration directory first to be able to go back if need it, or make a copy of the configuration directory to `pac.test` and run as:

`perl asbru-cm --config-dir=pac.test`

merikz:
> This breaks previous configurations.

So, is there some configuration migration in this PR? If so, have a look at #353.

hanspr:
Considering an interactive script, the original logic is very different from this one. I have not studied how to migrate older password entries to new ones yet. I never used the keepass integration because was hanging Absru, it took minutes to read mi file, and never liked the click right submenu with hundreds of duplicated entries, so I'm not very familiar with the current implementation.

hanspr:
@merikz 

I saw your implementation, its cool.

But I actually removed all the right click menus in my implementation, there are no KeePass submenus in PACMain, PACEdit or the terminal. The only right click menus left are inside edit/PAC...pm entries, I'll check but I'm really not that concerned on those sections and the code is very small.

hanspr:
@gfrenoy 

I think I have the basics covered for this one. I want to begin working on the migration script.

Where should I place this script?

Can I create a migration subfolder so we can place migration scripts in it?, name it "migrations"

Should we create a complete framework to handle configuration migrations or just stand alone scripts and run them in order with an entry script?

I think that the best place to check for migrations and run this scripts would be in asbru-cm start up script before we open the full application, but I would like to know how you would like this to be handled.

hanspr:
Added more functionality, now you can set the host as a reference to a keepassxc entry.

So the 3 more sensitive values in a connection will be only references in the configuration file.

imagen

gfrenoy:
> Where should I place this script?

In ```utils``` ?  A file named ```asbru_config_upgrade.pl``` ?

This script would always upgrade from the version found in the in configuration file to the latest version.

A parameter ```-f``` / ```--force``` will force starting from a given version.
A parameter ```-c``` / ```--config-dir``` will change the the path of the configuration directory

See also #353 

Thanks :)

skipperTux:
Hi. I have tested the latest commit 0cee5f9 on a Fedora 31 VM with KeePassXC 2.5.2 installed. I have two issues:

1. _Key file_ is not saved in Preferences when saving, while _Database file_ is. When I enter a key file, it shows in the field `Key file`, (no Master Password saved). When closing and restarting `asbru-cm` I cannot login because the key file is missing. If clicking `Cancel` to bypass KeePassXC login, the `Key file` field under Preferences is empty.
image

2. Search KeePassXC is working quite fine (although I think the Toolbutton is better located in the Authentication Panel). However when setting KeePassXC values for User and Password, the `Save and Close` button is not working anymore. The only option to close the dialog is `Close without saving`, what - obviously - does not save anything.
image

Any logs I can check or provide to see what is going on in more detail? I did not found anything in `/var/log` or using `journalctl -b -f`. What came up in `journalctl` (with delay, but it seemed related to clicking `Save and Close`):

```bash
Feb 07 17:02:55 localhost.localdomain gnome-shell[1832]: libinput error: client bug: timer event3 debounce short: offset negative (-0ms)
```

P.S.: I like the redesigned ui :smile:.

hanspr:
@skipperTux 

Thanks for the information and testing. I will check the Keyfile issue.

And test the close button, with something might have changed with so many merges that broke that part.

The location of the Search can not be relocated there because that row gets disabled if you change the auth method to privet key. But in the redesigned UI I will moved up together with the host:port. But I'm still playing around trying to place everything in better positions.

hanspr:
I fixed the issues reported, it should work.

And if you thing this ui change is nice, check PR #395 or my branch facelift-1

gfrenoy:
@hanspr How do we manage the upgrade of configuration files ?  Is there a risk to break the configuration and not being able to come back to a previous version ?

hanspr:
Hi @gfrenoy 

I have't written any script.

Going back?, only if you recover the previous backup, or we leave a special named back up, and downgrade.

The script would have to try to match the current password configuration against a unique path, like it does now.

The possible results would be:

* A match is found and the password is converted to the new standard
* Multiple matches were found, then the user would have to select the correct one
* No match is found. The user would have to configure the correct path manually on those connections
* At the end of the full process, we could display a summary and let the user decide.
    * Continue and manually fix what could not be converted
    * Go back to previous versions, we recover our latest back up, and abort. The user would have to downgrade.

We can ask this to the users that have been wanting a fix on this,  and find out their expectations and how much are or not willing to do in the 3rd scenario when a match can not be found.

---

Changing the subject, I found another unexpected event that I have to work on.

In the latests keepassxc version, If you change your database to the new standard (4.0) you can choose the level of encryption, and depending on the level, the database will unlock in as little as in 100ms, or as long as 5 seconds.

imagen

Because I'm now pulling the information on the fly, a 500ms or less will not be a huge impact, but if some paranoic user takes it to the top. 5 seconds to pull a username, or password, will be very slow. 
The impact will be visible at:

* Start up, where I check for the existence of a master password and check if it works
* Connections. Currently I pull the username and then the password on a second connection
* The process to test a password and failing.

So I have to figure out how to improve performance considering this new possibility.

The first one that comes to mind is to create a cache of the database (open, wait 5 seconds, dump the hole database and store in memory).

And that would bring a gain a lot of the original issues of : not seen updates on data changed in the database, pulling al unnecessary and unused entries, sneak into all their personal records that are not connection related and stored there.

I have though of some other less intrusive techniques, but I can go as low as the time for the database to unlock. In this case scenario 5 sec.

I guess that following the honour of the long tradition of options, I could add an option to catch the file and the user may decide if he is willing to wait the 5 seconds interactions. Or speed up things, understanding the consequences of this.


I hope I get some free time to test and send the changes next week.

gfrenoy:
1) About migration:

I understand:
 - There is no easy path forward
 - We don't have migration script
 - Once the configuration changed for 6.2, it cannot used with a previous version

What I still don't understand:
 - What we need to do to avoid that some accidentally loses all his/her configuration
 - What are the instructions we need to give to, at least, migration the configuration manually

2) About 4.0 (and the general "slowliness" of keepassxc-cli)

- Dumping the database in memory does not look the right thing to do ; there are some complex security thing behind this ;
- I would better see that we use the "open" keyword to open a shell and then interact with the database that is properly managed by KeePass itself ;
- Where did you see that interactive mode will be dropped ?  I still see active discussion about it (see [here](https://github.com/keepassxreboot/keepassxc/issues/3348)).

hanspr:
> What we need to do to avoid that some accidentally loses all his/her configuration

I really do not understand the accidental lose, nothing will be loss. In the worst of all case scenarios, you will have to edit the connection you need right now and write the username and password just like we currently do; and you will have your connection back again in a few seconds, nothing worst than that. Is this an inconvenience? Yes. But nobody will be without being able to continue working, that is for sure, no one will be locked out of his/her servers.

I think the warnings and of the migration script should explain what is about to happen. What the consequences will be and that if he/she needs further instructions go to the wiki pages.

> Where did you see that interactive mode will be dropped

Sorry, I got mislead by command name "open". Don't know why in my head thought the command was another.

Yes that should be the path, to open one interactive session (will take 5 seconds) and then I guess the rest would have no further delays. I will try to work on that.

gfrenoy:
> I really do not understand the accidental lose, nothing will be loss. 

Ok perfect :)  I was not sure so thanks for the confirmation.

> Yes that should be the path, to open one interactive session (will take 5 seconds) and then I guess the rest would have no further delays

Yes ; that'd be great.  See also my comment #415 ; I would not ask the password immediately at startup but only the first time it is needed.

> I will try to work on that.

Wonderful :sunglasses: !

hanspr:
Ask for master password on terminal open or request to list. Do not ask at start up.

Pull all values from the keepassxc entry on one connect to avoid multiple opens (faster response)

Cache the list of entries to accelerate search response. No sensible data at this point.

Moved keepassxc list button to the top, next to the host : port connections settings

Tested with encryption levels of : 100 ms to 5 seconds. Response time is good with an encryption level of 1.5 seconds or below.

glade file might have to be revised, the merges were very extensive.

gfrenoy:
I made some changes ; mostly cosmetic but the main point is to implement #415 (do not ask for password if not required).

I believe we are ready to merge this to loki ; aren't we ?

hanspr:
> mostly cosmetic but the main point is to implement #415 (do not ask for password if not required)

I think I already did this one, no?

What about this one? #414

gfrenoy:
>> #415

> I think I already did this one, no?

Not completely.  In PACTerminal, it was always asking for the password ; even if not need for that particular terminal.  Now it's checking if one of the field looks like a KeePass entry and only in that case will ask for the master password.

> What about this one? #414

Would be great but I did not do anything about this.  So far, I'm using the described workaround ;-p

hanspr:
>  it was always asking for the password ; even if not need for that particular terminal

My fear, is that I would have to scan everything (including expects, and any location where you can add this instruction) is reality is not limited to user and password.

skipperTux:
A few comments:

- #415 not a Perl guy, but shouldn't that be triggered when asbru needs to resolve fields with KeePassXC, and the KeePassXC module holds a state if the db is already unlocked?
- #414 provide a field in preferences to enter `/path/to/keepass/binary`. It defaults to what is found by asbru, but user can enter something else, like `/path/to/KeePassXC-*.AppImage cli "$@"`.
- Added the host with `<url|/...>` using `Add Url KeePassXC`, field is not cleared, url-value is added before existing IP; in contrast to User/Password fields are cleared before adding KeePassXC value.
- Added the host with `<url|/...>`, error opening connection `Can't call method "regexTransform" on an undefined value at /home/user/Projects/asbru-cm/lib/pac_conn line 410.` I tested with these values in KeePassXC url: http://IP, http://IP:PORT, IP, IP:PORT and cleared the field before, see above.
- How does the `<url|/...>` work with the port field? I can right click `Add Url KeePassXC`, but nothing happens. Where would the port be taken from? How would it be set in a numeric field?
- Feature: When `KeePassXC Search` is confirmed and User/Password is set, also set `Password Show` to true. The field does not contain a secret anymore, but I can see it is populated by KeePassXC.
- I never used to old KeePass integration, so I do not qualify for testing migration scripts to KeePassXC integration
---
Off-topic (I can move these to new single issues, if interest exist in resolving):
- WiFi icon for 'My Connections' is - IMHO - confusing
- asbru-cm crashes on xfreeRDP and Embed in Tab in Wayland, might as well be a xfreeRDB issue, related to #421. (I found this out by coincidence, because my test-VM is using Fedora 31 running Wayland.)

```bash
Gdk-Message: 18:08:15.930: Window 0x55d4fa6065d0 is a temporary window without parent, application will not be able to position it on screen.
Gdk-Message: 18:08:15.930: Window 0x55d4fa6065d0 is a temporary window without parent, application will not be able to position it on screen.

(asbru-cm:5563): Gdk-CRITICAL **: 18:08:15.949: gdk_wayland_window_handle_configure_popup: assertion 'impl->transient_for' failed

(asbru-cm:5563): Gtk-WARNING **: 18:08:29.726: GtkSocket: only works under X11

(asbru-cm:5563): Gdk-WARNING **: 18:08:29.727: gdkwindow-x11.c:5633 drawable is not a native X11 window
Segmentation fault (core dumped)
```

hanspr:
> it was always asking for the password ; even if not need for that particular terminal

Had added more detection points to validate if ask master password is necessary. I guess more points will pup up in time.

hanspr:
> Added the host with <url|/...> using Add Url KeePassXC, field is not cleared, url-value is added before existing IP; in contrast to User/Password fields are cleared before adding KeePassXC value.

Fixed, replace value for url, insert into cursor position for other text fields

hanspr:
>Added the host with <url|/...>, error opening connection Can't call method "regexTransform" on an undefined value at /home/user/Projects/asbru-cm/lib/pac_conn line 410.

Fixed regression of wrong merges

hanspr:
>How does the <url|/...> work with the port field? I can right click Add Url KeePassXC, but nothing happens. Where would the port be taken from? How would it be set in a numeric field?

Removed menu from text fields different from Host/IP

hanspr:
@skipperTux 

>Feature: When KeePassXC Search is confirmed and User/Password is set, also set Password Show to true. The field does not contain a secret anymore, but I can see it is populated by KeePassXC.

I do not understand this one, if you could help me with a set of images to follow the logic and understand the problem or feature request?

hanspr:
@skipperTux 

> I never used to old KeePass integration

me either :confused: , never convinced me.

But thanks very much for the time you have put into this testing and your comments, It has helped a lot.

skipperTux:
> @skipperTux
> 
> > Feature: When KeePassXC Search is confirmed and User/Password is set, also set Password Show to true. The field does not contain a secret anymore, but I can see it is populated by KeePassXC.
> 
> I do not understand this one, if you could help me with a set of images to follow the logic and understand the problem or feature request?

image

vs.

image

When the user selects KeePassXC value, and the Password field is cleared and populated with `<password|/...>`, set Show to true (once). I can see the field is populated with KeePassXC value, and if I want I can switch back to Show=False.

Well, thank you @hanspr @gfrenoy for your continuing work on asbru-cm!

hanspr:
So, if I understand you would like that Ásbru shows the extracted password value at the time of clicking into "Show", and not to show the mask.

Correct?

But the same concept could apply to username and host.

My premise was to show the mask used so you know you selected the correct one, and never extract the real values only at the time to connect.

This is also true for people that likes to use the tooltip option, it will not extract the user name or password. Mainly because it will be very slow to go one by one extracting the real values, and second, if some one is really that serious about hide information, will protect an unattended machine.

So, I understand the concept, and what is trying to achieve. Maybe I could do some mix of both, show the selected mask, and if show is selected, then show the user name or password in a tooltip when the user hovers over the text field or an icon. Will give it a thought to see how to achieve it.

skipperTux:
> So, if I understand you would like that Ásbru shows the extracted password value at the time of clicking into "Show", and not to show the mask.
> 
> Correct?
> 

No, misunderstanding, do not show the extracted password. My feature idea is only about the `Show` checkbox.

Auto-activate (once) the `Show` checkbox when KeePassXC mask is choosen from `KeePassXC Search`.

IMHO it does not make sense to hide the mask `<password|/...>`, it is helpful to see this field is populated by KeePassXC by showing the mask, not asterisks.

hanspr:
Ok, got it, thanks

Well let me see, again it should be visible only if a keepassxc mask entry exists, because people could have mixed configurations.

skipperTux:
I would set Show to true once when the user searches the KeePassXC database, chooses an entry and the Password field is populated with a KeePassXC mask. I would set `Show` to true when the KeePassXC module sets the value of the Password field with a mask. Just for convenience, maybe not worth the effort.

hanspr:
@gfrenoy 

I have tested this option for several weeks, and it works, does not produce any conflicts if you do not use KeePas.

Unless there is an impediment yet or we plan to drop it, could we merge it?

gfrenoy:
Yes, let's merge this into loki so that it's easier to merge etc.

There are a few more suggestions in this thread we may want to implement ; if not yet done, I'll create dedicated issues to have them all under the [6.2 milestone](https://github.com/asbru-cm/asbru-cm/milestone/3).

In any case, thanks a lot for this major improvement !

hanspr:
All the issues in the comments were fixed, the suggestion of showing the password can be implemented if but not that I really think is necessary to do. And the only one related to keeping the database unlocked was already discussed. So all has been implemented.

PR #350, opened 2020-01-08
fix343

### LINKED ISSUE #343: Keepass does not work in latest

hanspr:
Patch to control receiving undefined on emtpy values, fixes #343 

PACKeePass will be rewritten in loki, so no need to patch in loki.

Indentation fixed.

### DIFF
--- a/lib/PACKeePass.pm
+++ b/lib/PACKeePass.pm
@@ -191,42 +191,42 @@ sub _buildKeePassGUI {
     # Build a vbox
     $w{vbox} = Gtk3::VBox->new(0, 0);
 
-        $w{cbUseKeePass} = Gtk3::CheckButton->new('Use KeePassX');
-        $w{vbox}->pack_start($w{cbUseKeePass}, 0, 1, 0);
+    $w{cbUseKeePass} = Gtk3::CheckButton->new('Use KeePassX');
+    $w{vbox}->pack_start($w{cbUseKeePass}, 0, 1, 0);
 
-        $w{cbKeePassAskUser} = Gtk3::CheckButton->new('Ask user when multiple matches are found');
-        $w{vbox}->pack_start($w{cbKeePassAskUser}, 0, 1, 0);
+    $w{cbKeePassAskUser} = Gtk3::CheckButton->new('Ask user when multiple matches are found');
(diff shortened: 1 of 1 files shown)


gfrenoy:
Hans, "hiding" the issue is fine but how do we explain that is was not an issue with v5 and that the same file read with the same "decoder" is now causing an issue with v6 ?

hanspr:
Well the issue is somewhere KeePass.pm

I'm not really that familiar with that module, and is strange that it works in my system, yours, there are no more reports from other people.

So it sounds again as an issue related to environment of versions and compatibilities between libraries in his computer (perl, encryption libraries, database version, etc).

If it is a combination of libraries that produces this, we first would need to setup an environment that allows us to debug, if we can not get it we would not be able to move on.

And finally, the new integration does not depend on that module. I see this as a quick patch while we move on.

gfrenoy:
Ok well, let's hide the problem for now and let's focus on the new way to use data from a KeePass file.

PR #356, opened 2020-01-12
Fix67 1

hanspr:
In reference to comments on #67

> width of the port field should be increased (to allow 5 digits)

Increased

>when no private key is selected, option IdentityFile remains empty in temporary ssh_config file (which causes an error)

IdentityFile for destination server is now optional and not included in ssh_config

Jump server IdentityFile is obligatory will die informing the user that is required

>As we are using the -F option of ssh, the system-wide file will be ignored ; either we just document this or maybe we can try to include them in the on-fly-generated config file ?

Was already merged and solved before on #348

>it always use the field Private Key > User in Connection Details for the connection to main host ; especially if that field is empty, it will fail ; I believe we should check if User / Password is checked and use that if given (with PreferredAuthentications to password)

IdentityFile for destination server is now optional and not included in ssh_config

>Jump option is specific to SSH so it should not be displayed for other methods (RDP or VNC, typically)

I leave this one to you

> Check of SSH should only happen for SSH method

Do not understand this comment


gfrenoy:
Thanks a lot :-)

>> Check of SSH should only happen for SSH method
> Do not understand this comment

So for RDP or VNC connection (or any other "non-SSH" method) there's no need to check for the SSH version since the option will be hidden anyway.

gfrenoy:
> Jump server IdentityFile is obligatory will die informing the user that is required

This is not required.  It really depends on how your ```ssh_config``` is defined ; as a matter of fact, it just works fine in my case (without any ```IdentityFile``` defined) because it will reuse my default identity that is accepted by the jump host.

I would simply leave it empty and let it to the user.

hanspr:
Got it, but the intention of the check is to know in advance the capabilities of SSH, I would not know if the user will use ssh or not, or change its mind every 2 seconds, or play around with the checkbox. And who knows if `-J` option is not available and decides to upgrade, then the option would be available.

Does it really complicates things checking?

hanspr:
> Jump server IdentityFile is obligatory will die informing the user that is required

 Give me a second and will make it optional then

hanspr:
> Jump server IdentityFile is obligatory will die informing the user that is required

Now is optional

gfrenoy:
>  I would not know if the user will use ssh or not, or change its mind every 2 seconds,

Mmm, yes this could be "dynamic", indeed.  Not only whet you start the edit dialog.

> Does it really complicates things checking?

Well, the reasoning is more to just do what is strictly necessary and make sure we don't have dependency where should not be.

But let's try like this ; we'll see.

Thanks again !

gfrenoy:
Ok I believe I fixed most of the issues I found.  I also re-arrange the glade file and greyed out the Jump options (as it was done for the Socks proxy).  Let me know if it looks good for you.

I had to remove some ugly "show_all" that were added by David to make sure the new specific options were shown.  It will deserve more tests I believe...

One more to go (but too late for today) : I could not "remove" the private key I had previously selected.  I did it by editing the config file manually ; not nice ;)

I've added a "Clear" button but there's no code behind it yet.

PR #357, opened 2020-01-12
Fix67 2

hanspr:
>I've added a "Clear" button but there's no code behind it yet.
>One more to go (but too late for today)

hand-off 

Finished clear button code, added another to private key in connection configuration

Fixed clearing the chooser button without warnings.


gfrenoy:
Thanks :+1:

PR #362, opened 2020-01-17
Fix359

### LINKED ISSUE #359: ssh compression activated by change of other options
**Describe the bug**
When adding a Dynamic Socks Proxy (by clicking the "Add" button) for a connection, the "Use Compression" option is automatically activated.

**To Reproduce**
1. Create a new connection by clicking the "New CONNECTION" button in the connection tree panel
2. In the "Edit Connection" window -> "SSH Options": Disable the "Use Compression" option
3. In the "Dynamic Socks Proxy" tab: Click the "Add" button.
4. The "Use Compression" option is erroneously enabled again.

The same thing happens when clicking "Add" in any of the tabs, "Local Port forward", "Remote port forward", "Dynamic socks proxy" or "Advanced Options". "Use compression" always gets enabled.

**Expected behavior**
The "Use Compression" option should only change when the user clicks the check box.

**Screenshots**

 - Ubuntu 18.04.3 Cinnamon 
 - asbru-cm v6.0.2

hanspr:
Fix #359  and improvement on port enumeration if used clicks multiple times


lib/method/PACMethod_ssh.pm
```diff
@@ -623,8 +623,9 @@ sub _buildGUI
     $w{btnadd}->signal_connect('clicked', sub {
         my $local_port = 1;
         $$self{cfg} = $self->get_cfg();
-        if ($$self{cfg} =~ /L.+(\d+):/) {
-            $local_port = $1+1;
+        my $next = () = $$self{cfg} =~ /-[LR]/g;
+        if ($next) {
+            $local_port = $next+1;
         }
```
merikz: This leads to "crash" of forward handling (all forwardings lost) at 3rd click on "add", if the first forwarding is with local port 2, since the $next value already exists. 
Message on console:
GetOptionsFromString: Excess data "CONFIG ERROR: Local Port 2 on Local IP  defined in Local Port Forwarding is already in use!" in string "CONFIG ERROR: Local Port 2 on Local IP '' defined in Local Port Forwarding is already in use!"

A way to find a safe default value is to take the highest existing value and add one, something like this:
```perl
@@ -623,10 +623,10 @@ sub _buildGUI
     $w{btnadd}->signal_connect('clicked', sub {
         my $local_port = 1;
         $$self{cfg} = $self->get_cfg();
-        if ($$self{cfg} =~ /L.+(\d+):/) {
-            $local_port = $1+1;
-        }
         my $opt_hash = _parseCfgToOptions($$self{cfg});
+        use List::Util qw(max);
+        my @usedPorts = map {$_->{localPort}} @{$opt_hash->{forwardPort}};
+        my $local_port = max(@usedPorts) + 1 if @usedPorts;
         push(@{$$opt_hash{forwardPort}}, {'localIP' => '', 'localPort' => $local_port, 'remoteIP' => 'localhost', 'remotePort' => 1});
         $$self{cfg} = _parseOptionsToCfg($opt_hash);
         $self->update($$self{cfg});
```
This goes of course for the remote and dynamic forwarding as well.

hanspr: List::Util

Is a standard library? I would not like to introduce more dependencies

merikz: "List::Util was only added to core in 5.7 (a development version) and the first stable release of perl containing List::Util was 5.8.0."
([From here](https://stackoverflow.com/questions/8694638/isnt-listutil-part-of-standard-perl-distribution/8694656))

hanspr: After testing your solution still has issues with the interaction with remote ports, you would have to scan both hashes forwardPort and remotePort, to actually know the maximum.

Right now, I'm trying to control that if you create 2+ entries one after another there is no crash for a simple interaction like that.

And even then the crash can happen at any time, if the user repeats local ports inside local and remote forward.

The actual solution would be to always check for non repeating values on every user interaction inside the GUI and warn, maybe on a future round.

merikz: Why do you bother about remote and local forwards at the same time? 
The listening ports (that must not clash with each other) are on different servers, local node vs remote node.

Yes, I agree. Best would be to verify the user input to be valid.

But the UX of this is not very good. You click "Add" and an entry is added, but it's not immediately visibly which one is the new one. 
A future round, as you said.

hanspr: >Why do you bother about remote and local forwards at the same time?

imagen

If the ports repeat in those points it will crash as you described earlier.

merikz: hmmm... that's a bug.
It should be possible to have several remote forwarding to a local service. Say that I run a webserver on port 80 locally. I should be able to have several ports on the remote node that leads back to the webserver.
Also, I can have many webservers at my local site on different local servers all running on port 80.

hanspr: >hmmm... that's a bug.

yes it is, if have some time left will try to find and remove it on this PR.

hanspr: Ok, have pushed a new PR, same logic for looking for the maximum port (no dependency, not as elegant as yours but works).

And removed validation for repeated port comparing local port in remote forward.

Git it a check, hope is more stable.

hanspr: Or, maybe we should let a users configure ports as he wants and not rise any errors on duplicated ports, it will come out when he launches the connection.

merikz: For a Remote Port Forwarding, it is the port on the machine you are connected to that must not repeat. This is from ssh point of view.

The asbru UX here is bad.
The meaning of "remote" and "local" is reversed in a Remote Port Forwarding compared to the Local Port Forwarding. 
For a Local Port Forwarding, the "Bind Address" and "Local Port" refers to your local laptop, and the "Remote Address" and "Remote Port" refers to the environment of the machine you connect to.
For a Remote Port Forwarding, the "Bind Address" and "Local Port" is the address of an interface and port on the machine you connect to with this asbru connection (the "Host"), and the "Remote Address" and "Remote Port" is something on the local network of the laptop (or the laptop itself in the case of "localhost").
This is weird! 
 
So, for our discussion about the code, it is correct that the $local_port must not repeat. It's just that what is "local" at a remote forward is not on the same machine as the "local" for the local forward. So they can be same, they wouldn't clash from ssh point of view. 
If asbru cannot handle that, asbru is wrong.

lib/method/PACMethod_ssh.pm
```diff
@@ -623,10 +620,20 @@ sub _buildGUI
     # Button(s) callback(s)
     $w{btnadd}->signal_connect('clicked', sub {
         my $local_port = 1;
+        my @usedPorts = ();
+
         $$self{cfg} = $self->get_cfg();
-        my $next = () = $$self{cfg} =~ /-[LR]/g;
-        if ($next) {
-            $local_port = $next+1;
+        my @ports = split /-[LR]/,$$self{cfg};
```
merikz: @ports would include parameters that aren't ports, right? Stuff like -A -X -C from "other options". Isn't that a problem?

lib/method/PACMethod_ssh.pm
merikz: > Or, maybe we should let a users configure ports as he wants and not rise any errors on duplicated ports, it will come out when he launches the connection.

UX: I think we should have a set of (empty) fields to fill in with an "add" button beside them. No default values.
Then perhaps do not check the values more than simple sanitation, like numbers for port, no spaces etc. No cross checking between entries.

But before deciding, maybe double check what it looks like when the user misconfigures it. Will openssh give reasonable hints?

lib/method/PACMethod_ssh.pm
hanspr: Asbru is putting them at the beginning so they are discarded.

lib/method/PACMethod_ssh.pm
hanspr: Ok, lets see if we can round up all this

* Removed all the crashing messages and substituted for message warnings on the UI
* The way all this is coded originally does not allow to have blank ports, so the enumeration remains
* Your implementation has been placed for locating the maximum port

>For a Remote Port Forwarding, the "Bind Address" and "Local Port" is the address of an interface and port on the machine you connect to with this asbru connection (the "Host"), and the "Remote Address" and "Remote Port" is something on the local network of the laptop (or the laptop itself in the case of "localhost").

I just need to make sure if what you say is that the place of this 2 labels should be reversed for Remote Port Forwarding

imagen

Or I leave it as it is now?

I ask, because I'm not a expert ssh user, I use the basics.

lib/method/PACMethod_ssh.pm
hanspr: Added your implementation

gfrenoy:
Great additions !  Thanks both for your help and contribution :)

lib/method/PACMethod_ssh.pm
merikz: Yes, that is what I'm saying. 
But then perhaps we should use the wording openssh has in the man-page, and stick to that. The whole concept of remote forwarding is already confusing as it is for the non-experts, so sticking to the concepts from the manual would make it easier for the user to google if he don't understand.

My `man ssh` (from ubuntu 18.04.3) calls them this:
-L [bind_address:]port:host:hostport
-R [bind_address:]port:host:hostport

But since this PR has been merged already, I guess we can take this in another round.

PR #367, opened 2020-01-18
Moveproxyssh

hanspr:
SSH Proxy Jump Host set up may be like this?


hanspr:
I think this could help to build proxy,jump options depending on each method.

Currently enabled for ssh and sftp that I think are the ones that need it.

gfrenoy:
Good to have this a bit more factorized but the main issue is how to make this clear for the end user...

Today, we have 2 types of proxy: SOCKS & Jump Server.  And we can set this at the global level (in the preferences) and override this for each connection.

Jump Server is only for SSH connections.  This option is not shown in the connection settings.

SOCKS is today shown for all connections.  But which one do really support that ?

hanspr:
>Today, we have 2 types of proxy: SOCKS & Jump Server. And we can set this at the global level (in the preferences) and override this for each connection.

I think that there are a lot of this issues all around the connections settings not only for ssh. Just to mentions some of them.

* ssh port forwarding, normal, reverse, dynamic, jump server.
    * There must be a lot of incompatible combinations that will not work. Or maybe they work but the order in which they must be written is not clear to me at least.
   * There are new features arround ssh that would be interesting, like setup a connection that opens tunnels and binds ports, and be able to relate it to other connections that need that one to be run first to work. And open the first hidden by default but not to duplicate if is already running, when more than one terminal has it configured. (hope did not confuse you with this description).
* rdesktop (has not been updated is looking for a new mantainer : https://groups.google.com/forum/#!topic/rdesktop-announce/AddglSNxK90), I think should be removed, as an option and only use xfreerdp.
*  vncviewer settings have changed(already working on this) and again some options are incompatible with in the same vncviewer settings.

So, you should open a new or multiple issues and:

* Put a list with your plans.
    * The first point should be to stop and do a full review of the configuration windows and plan how you want it to work and how you want it to look. For example:
        * Remove global proxy, jump options for good?, I bet there is already one person behind a firewall that will say "no".
        * Move jump server to ssh options specifically so we do not have to control the enable disable?
        * Or if set and the connection is of a different type (open a second hidden terminal to create that tunnel?
    * Invite someone from the group with good knowledge of ssh options, to build a readable matrix of combinations that can or not work together.
* Added them to your next milestone

What I'm trying to say is that the matrix of options and the possibilities of how to handle them is large. And it deserves a stop, analyse, discuss and make some agreements. 


>Jump Server is only for SSH connections. This option is not shown in the connection settings.

Don't follow, I see it in the Global and Network Settings Tab as we agreed based on the original proposal

gfrenoy:
Of course the full picture would be better but let's try to focus on the new things for 6.1 and ensure they make sense.

For SSH, it works well ; I use most of those options:  port forwarding, normal, reverse, dynamic, and now also the jump server etc.  Up to now, it works fine in all the combinations I'm using.  So don't fix it :)

For the other connection types, my question is : do the proxy settings (global or local) work ?

As far as I can see ; they do not.

So my thinking right now is that we should maybe just make clear that the network settings are only of SSH today.

And, as you suggest, take the time later on to review the other connection types...

hanspr:
> do the proxy settings (global or local) work

I can no assert for the global options

But I'm sure the local ones work, I tested. But than again I might be doing it wrong, here is my tests.

* First connection to a visible server and open a Dynamic Proxy on local por 5000. This could be done with any other tool, or loaded in a bash script at start up, who knows. For testing purposes I did it with Asbru.

imagen

imagen

* Second connection, connects to the local port 5000 to access a 2nd machine behind the remote firewall

imagen

imagen

imagen

I don't know how do you test? How should it be done then?

gfrenoy:
My question was about the other connection types (so not for SSH)...

For VNC, RDP, generic, others, ... Do we hide the proxy options or just keep them as they are (ie: not working...) ?

hanspr:
My first question would be: Do you know if it was working before the changes?

If it was, then we should find what changed that it broke it.

If you do not know, or it never worked before.

Ten, leave it as it is, it has been like that since the original application and like many other things will have to be fixed, but we have not changed anything really.

As a side comment, in the case of vnc options : has this vague option "Connect via", where apparently you write your ip:port proxy server.

hanspr:
> As a side comment, in the case of vnc options : has this vague option "Connect via", where apparently you write your ip:port proxy server.

What I do is:

* Connection one opens SSH with  remote server, and creates reverse port forward local:5900 <-- remote:5900
* Second connection VNC to local:9500

So there is a current work around, for connecting to remote vnc, rdp using normal ssh configurations.

gfrenoy:
> My first question would be: Do you know if it was working before the changes?

Good question yes.  I would say it had probably worked but I never tested this myself.

> what changed that it broke it.

Net::Proxy has been removed.  This was supposed to work for all types of connections.
(but I could never make it working ;-p)

> Ten, leave it as it is, (...)

Yes but is there any "quick things" we can do to make it clear that it will only work for SSH ?

I changed the tooptips already for Jump server.  Maybe we should put the same for SOCKS ?

hanspr:
>Net::Proxy has been removed.

So, it was used for vnc, rdp. Did not realize that, because I always used other techniques  to accomplish that.

I would suggest then that we review and modify #368 and find a way to bring it back with what we have.

* Test the vnc via option and if i works make it clearer
* RDP, as mentioned before, perhaps opening 2 connections internally
    * first connection hidden doing something like `ssh -L 13389:[Windows Server RDP address]:3389 [address ssh server] -l [ssh username] -N`
    * second connection opens xfreerdp to localport:13389
    * add this configuration to rdp options

That is what it comes to mi mind right now

hanspr:
Sorry if I'm slow with this stuff,

I check the vnc option : Open via and found this message

imagen

Apparently the technique is the same always

1. Open an ssh connection to some server, open port forwarding.
2. Second connection open the rdp vnc session lo localhost

As proposed earlier in my large comment

>There are new features arround ssh that would be interesting, like setup a connection that opens tunnels and binds ports, and be able to relate it to other connections that need that one to be run first to work. And open the first hidden by default but not to duplicate if is already running, when more than one terminal has it configured. (hope did not confuse you with this description).

This could create a very flexible common technique for all connection types

1. Create connection "A" that opens a tunnel to a remote server.  So we have all the ssh options available again : ip,port,user:password or keys connections, compression, xforward, etc, etc, etc.
    1. Add another configuration settings to list of proxy tunnels
2. Connection "B" has the option to link this connection on the "Network setting" tab
    1. No proxy
    2. Global proxy
    3. Use Tunnel from (Select tunnel connections configured), in this case "A"), option to open this connection (hidden or visible).
    4. Proxy (for ssh only)
    5. Jump Host  (for ssh only)

imagen

3. Configure the vnc,rdp,ssh,etc connection assuming the linked connection "A" will be started before (opening the tunnel).

Does it makes sense? or is to over-engineered?

gfrenoy:
Very good finding for VNC !

Maybe we can remove that "via" option from the specific VNC settings and re-use what we called "network settings".

So, at the end of the day, we can probably find ways to make this the same for every connection type.  That will probably take time to design this properly... You idea of reusing an existing connection is also interesting.

But we should probably have less ambition for 6.1 ; don't you think ?

What if we just say that the next proxy settings only work for SSH in 6.1 and let's try to write down some new use cases / issues for 6.2 (or 6.3 or later...) ?

hanspr:
Fine by me.

PR #368, opened 2020-01-19
Rdpvnc

hanspr:
As a representative of the non tabbed users community (AKA gimp like style users), I send this PR : 

* Opening vnc,rdp connections requires a helper window connection to open the one that we want. And is left open during the session.
    * This changes hide that window and brings it back to the front only if a connection error occurs, other wise remains hidden.
* I made a revision of vncserver settings to make it work again with vncviewer, is tested and working. And had to redesign the Options tab, to handle the new parameters.


hanspr:
@gfrenoy 

Did Klass, changed his username like bitlord, or this validation is broken?

gfrenoy:
I think he has several accounts, depending on where he connects from...But don't worry, this does not prevent me from merging.   I'll have a look ; I think I'll need to first merge back from master before merging this one.

KlaasT:
Ist really depends If I am working from office or home. Different E-Mail addresses. As said that won't prevent from merging :-)

PR #370, opened 2020-01-20
Portforward improvement

hanspr:
In my last PR #362 

Had a conversation with @merikz , and could not finish the suggested changes, that I think now understand.

There were some issues left, unsolved.

1. We have 2 tabs port forwarding and 1 for dynamic proxy. In the 3 of them some port will have to be bind in the local computer. Those local ports can not be repeated on the 3 tabs. So validations and an enumeration has to work for the 3 tabs. (marked in red on the images)

2. The add button will not work correctly if the IP fields are left blank, so I added a default 0.0.0.0 to remote address. And a default localhost to the local address. This should help even more how to setup the fields correctly.

3. The ssh manual and the labels were confusing for the Remote port forward. I think , that the original intention of the labels, was to preserve the order in which this values are printed in the  ssh manual.

`-[LR] bind_address:port:host:hostport`

In the case of -L `bind_address:port` represent the local computer

In the case of -R `bind_address:port` represent the remote computer

So to make things clearer and better documented in the tooltips, I swapped the labels for Remote port forward, and edited a (hopefully) clearer tooltip, to so the configurations look like this.

imagen

imagen

imagen

This should cover the validations and configurations better.


merikz:
> In the 3 of them some port will have to be bind in the local computer.

This is not correct for remote forwards.
(I'm using the terms _bindaddress_, _port_, _host_ and _hostport_ from the ssh manual in the following.)

The _hostport_ for a remote forwarding does not clash with, and should not be validated against the _port_ (nor _hostport_) for the local forwardings.

**For a remote forward**, what you do is to tell what should happen if someone sets up a connection to the remote machine on the _port_. It should be tunneled through the ssh connection back to  the local ssh client that will forward the connection to _host_:_hostport_. This means:
- _host_: this can be any ip address or hostname on internet. No restrictions at all (except 0.0.0.0 which is not a routable address). It's common to use 'localhost', e.g. if you run a server on the local laptop, but could be anything. 'localhost' is, however, a good default value for a new entry.
- _hostport_: This is a port on _host_. Asbru cannot know anything about this, except that it needs to be a number 1-65535 (as we're not even considering anything else than TCP tunnels, for now :-)). **It does not need to be unique.**
- _bindaddress_: This is an address on the remote machine running the ssh daemon this asbru connection connects to. It indicates an interface on the remote machine from which the ssh daemon will accept incoming connections to send through the tunnel. **This can be empty, which should be the default value** as that is the default value the ssh client uses.
Using 0.0.0.0 here is not the same as empty. (See GatewayPorts in the sshd_config manual for details.)
- _port_: The port on the remote machine that the ssh daemon will listen on. Asbru cannot know anything about this, except that **it is wrong to use the same value on several entries**, as the ssh daemon binds to that port. Same validation and default values as for local forwardings can be used, but they don't clash with each other. 

As for local forwardings and dynamic socks proxies; I agree, they cannot be the same as local forwardings since they both bind to the local computer.

PR #371, opened 2020-01-20
Portforward2

hanspr:
@merikz 

Thanks one more time.

Hope I got it right this time, the language of this network stuff is not so easy to understand for me.

* Port error message between : local forward and dynamic proxy : remains
* Port warning message of remote forward: only for repeated bind_address:port inside remote forward, not compared against local or dynamic proxy
* Removed 0.0.0.0 in remote forward


gfrenoy:
Thanks to both of you !  The overall quality is increasing :)

Issue #372, opened 2020-01-21
Asbru 6.0.2 does not save text in 'info tab'


thijz:
**Describe the bug**
When you add comment in the info tab of a connection this text is not saved.

**To Reproduce**
1) open a connection
2) goto the info tab and enter some text
3) switch to the connection tab
4) switch back to the info tab : text is gone

**Expected behavior**
IIRC there used to be a 'save' button but that is no longer there (i assume Asbru should be auto saving the text on the info tab?)

**Screenshots**

**Desktop (please complete the following information):**
 - Ubuntu 19.10
 - Gnome 3.34.2

**Additional context**

hanspr:
The info tab is a notification area from the original application, and is __not__ meant to write on it or save anything at all. Should not even allow to type anything in it, to avoid creating a confusion like yours.

Perhaps you are looking more for a new feature, I would suggest that you think of what you are trying to accomplish with that action and request a new feature, and no necessarily based on editing the info tab, maybe adding "notes" to he connection settings and display them in the info tab when e connection is selected?.

popxunga:
Can't we block the keyboard input on Info tab to avoid this ?

thijz:
@hanspr 
Really ?  before switching to asbru (when i was still using PAC) i used the info tab all the time. it was really handy to have a place to add some extra info about the connection.
When i switches to Asbru i only noticed after a couple of days that all this info was gone. lost forever :-(
If the text in the info tab is not meant to be edited and if it does not get saved (it definitely did in PAC!) then there is indeed no reason to make it editable.
However AFAIK the only info that is ever in the info tab is the connection name.  not much added value in that IMHO.  i cant imagine that the info tab was ever added just to show the connection name ...

Would it be possible to somehow split the info tab in a RO section (for the connection name ;-) and an editable section ?  To me It just seems like the obvious place to add info about the connection is in the 'info' tab.  Dont you agree ?  :-)

gfrenoy:
I see indeed some lines in the code that make thinking it should work as you describe:

https://github.com/asbru-cm/asbru-cm/blob/master/lib/PACMain.pm#L2169-L2182

If was not intentional to remove that feature ; just that we did not even know it was working like that ;)  I'll have a look and try to restore this.

Thanks for the heads-up.

thijz:
thanks @gfrenoy that would be fantastic !!
oh and while you are working on it could you also add the possibility to paste screenshots there ? (and video's and flash animations  ;-D

[...]

PR #376, opened 2020-01-23
Fix372

hanspr:
Remove edit and focus on info tab to avoid confusion that is editable and related to the connection as in #372


KlaasT:
Thanks for your help on that issue.

PR #380, opened 2020-01-24
retry on timeout

### LINKED ISSUE #373: Some servers disconnected when using clusters
Hi,

Thanks for the great software! When I want to login on multiple servers when using clusters (say 15 - 25 servers at once), then most of the servers are connection correctly the first time. Some of the servers are disconnected and I need to reconnect them manually. Is there a possibility that when a server is disconnected when connected the firstime, it will try again? I think this issue occurs when connecting to too much servers at one.

Smikanl

hanspr:
Fix #373 back ported to master, just the minimum need it for this to work on main branch.


KlaasT:
Thanks very much.

Issue #297, opened 2019-12-01
Support for gtk-vnc


gfrenoy:
[gtk-vnc](https://wiki.gnome.org/Projects/gtk-vnc) is a a VNC viewer widget for GTK that can provide a "proper" GTK-style embedding solution for VNC connections.

Debian ships these files in its [gir1.2-gtk-2.0](https://packages.ubuntu.com/xenial/gir1.2-gtk-vnc-2.0) package for a long time so it should be widely supported.

And example is available in the gtk-vnc project:
https://gitlab.gnome.org/GNOME/gtk-vnc/blob/master/examples/gvncviewer.pl

popxunga:
Unless is done using ssh tunneling, VNC is highly insecure and it's usage not recommended.

gfrenoy:
>  VNC is highly insecure and it's usage not recommended.

Well, as any tool, you have to use it wisely.

figue:
Nice idea. In Archlinux it's available in official repos:
```
figue@pluto ~ % LC_ALL=C pacman -Si gtk-vnc
Repository      : extra
Name            : gtk-vnc
Version         : 1.0.0-1
Description     : A VNC viewer widget for GTK+
Architecture    : x86_64
URL             : https://wiki.gnome.org/Projects/gtk-vnc
Licenses        : LGPL
Groups          : None
Provides        : None
Depends On      : libgcrypt  gtk3  gnutls  libpulse
Optional Deps   : None
Conflicts With  : None
Replaces        : None
Download Size   : 137.03 KiB
Installed Size  : 879.00 KiB
Packager        : Jan Alexander Steffens (heftig) <jan.steffens@gmail.com>
Build Date      : Wed Aug 7 18:27:55 2019
Validated By    : MD5 Sum  SHA-256 Sum  Signature
```

Mjolinir:
Looks like it is also in the Fedora repositories:

````
Last metadata expiration check: 1:30:13 ago on Thu 02 Jan 2020 06:09:34 AM EST.
================================== Name & Summary Matched: gtk-vnc ===================================
gtk-vnc2-devel.i686 : Development files to build GTK3 applications with gtk-vnc
gtk-vnc2-devel.x86_64 : Development files to build GTK3 applications with gtk-vnc
======================================= Name Matched: gtk-vnc ========================================
gtk-vnc2.x86_64 : A GTK3 widget for VNC clients
gtk-vnc2.i686 : A GTK3 widget for VNC clients
````

PR #381, opened 2020-01-25
Vncviewer 1

hanspr:
vncviewer in the latest version has changed how it operates. I do not know at what version the behaviour of vncviewer changed, had never used it before in asbru (PAC).

I have no way to test with older versions.

Currently there is no output to run expect, so in asbru, dies with timeout all the time. Because connection authentication is window driven, like this

imagen

This is a small workaroundthat should not affect older versions of vncviewer, but ignores timeout so newer version does not dies.

I do not know if we should begin testing vnc version available and decide to allow or not the timeout expect.


gfrenoy:
Which version of vncviewer are you talking about ?

hanspr:
```
VNC(R) Viewer 6.19.923 (r41938) x64 (Sep 19 2019 16:09:33)
Copyright (C) 2002-2019 RealVNC Ltd.
RealVNC y VNC son marcas comerciales de RealVNC Ltd y están protegidas por los
registros de marcas y/o por solicitudes de registro en trámite en la Unión
Europea, los EE. UU. de América y otras jurisdicciones.
Protegido por la patente del Reino Unido 2481870, la patente de Estados Unidos
8760366 y la patente de la UE 2652951.
Consulte https://www.realvnc.com para obtener información sobre VNC.
Para ver los reconocimientos de terceros, consulte:
https://www.realvnc.com/docs/6/foss.html
OS: Linux Mint 19.3, Linux 5.3.0, amd64

```

hanspr:
I guess I have annoyed Mr Travis

hanspr:
Ok, I know what is going on. While working with the timeout to fix the previous issue on reconnect cluster terminals, I reduced the connection timeout to 2 seconds.

So the bug to Real vncviewer was hidden on the fact that 0 timeout wait is infinite. And that is why the symptom was not showing up before.

So I have added the logic to skip the timeout close event on  Real vncviewer that is the one affected just in case someone sets the timeout parameter.

gfrenoy:
Can you add note in the code to remember why we're having a special case of RealVNC ?

And I think we should be even more specific on the tests because if this is not "Tiger", it does not mean we are using RealVNC !  As a matter of fact, the two supported VNC clients are [TigerVNC](https://tigervnc.org/) and [TightVNC](https://tightvnc.com/)...

We can make a quick fix so that it works for RealVNC but let's make sure we don't break the other clients and let's move quickly to [Gtk-VNC](https://github.com/asbru-cm/asbru-cm/issues/297)...

hanspr:
Done

hanspr:
Please put this merge in stand by, have discovered I need more changes

hanspr:
This complicated a lot, but I think I got it.

Considering the 3 possible versions of VNC, the configuration interface had to handle valid options, and names that are different.

Additionally the command line is the only place to store options. So the vnc client name had to be placed there and removed later at the time of connection.

hanspr:
merged with #384 , will close, not worth it to have 2

PR #384, opened 2020-01-27
Proxyvncrdp1

hanspr:
Round 1.

This PR has all required changes

* SOCKS proxy hidden in non ssh sessions
* Reuse Jump Host option for : ssh , vnc, rdp. Hidden in other types

__RDP/VNC View__

imagen

imagen

__SSH__

imagen

__Others__

imagen

* pac_conn opens a tunnel when requiered using

`ssh -F $CFG_DIR/tmp/$UUID.conf -S $CFG_DIR/tmp/$UUID.ctl -f -N -T -M -L $PORT:localhost:$PORT jumpserver`

* Closes the tunnel when connection ends or error.

`ssh -F $CFG_DIR/tmp/$UUID.conf -S $CFG_DIR/tmp/$UUID.ctl -O exit jumpserver`

* Tests
    * Only with Real vncserver
    * The principle is the same so I would expect if works for other vnc clients and rdp

---

From the TigerVNC Manual, they all solve it like this

imagen


hanspr:
PR #381 is already merged with this one.

hanspr:
It took me a long while to understand but, this is the good one:

* Detect vncviewer installed: Tiger, Real, Other
* Created separated method class for RealVNC
* Recovered original PACMethod_vncviewer.pm
* Added label to identify which one was loaded
* Removed via, from Tiger and vncviewer because will not be used. Will use this implementation.

lib/pac_conn
```diff
@@ -904,11 +904,39 @@ sub _setSshProxy {
...
+            # Check if port we want to bind is not alreay in use
+            while (($break)&&(my $msg = `netstat -lnt | grep $LPORT 2>&1`)) {
+                if ($msg !~ /:$LPORT/) {
+                    # netstat not installed?
+                    last;
+                }
+                $LPORT++;
+                $break--;
+            }
+            system "ssh -F $CFG_DIR/tmp/$UUID.conf -S $CFG_DIR/tmp/$UUID.ctl -f -N -T -M -L $LPORT:localhost:$PORT jumpserver";
```
gfrenoy: Should be $IP instead of localhost.

You are making a tunnel that starts on local port ```$LPORT``` of your machine to destination port ```$IP:$PORT``` of the initial target ; then you'll need to use ```127.0.0.1:$LPORT``` to actually connect.

hanspr: Sure?, because it works as it is,

is not that localhost is where the remote IP where the PORT should be transmited?

So is from my localhost --> his localhost:PORT?

And I could say my localhost:$LPORT --> 127.0.1.10:$PORT? saying that I want to send my LPORT to his port but in that address?

Just asking

hanspr: But I understand that maybe (do not know) if the other machine is not correctly configured and can not resolve local host, the safest way would be LPORT:127.0.0.1:PORT

Is that what you mean?

gfrenoy: Yes sure.  It will only work if your target is the same as your SSH gateway but if you are using A as connection target and B as SSH gateway, it will not work.

hanspr: Sorry Robber Duck, think I got it

gfrenoy: Np, one of my [favorite article](https://blog.codinghorror.com/rubber-duck-problem-solving/), btw :-)

lib/pac_conn
```diff
@@ -1090,9 +1120,18 @@ if (defined $METHOD) {
...
+        my $method = 'VNC';
         $CONNECT_OPTS =~ s/\s*-embed//go;
-        if (`vncviewer --help 2>&1 | /bin/grep TigerVNC`) {
+        if (`vncviewer --help 2>&1 | /bin/grep RealVNC`) {
+            $method = 'RealVNC';
+        } elsif (`vncviewer --help 2>&1 | /bin/grep TigerVNC`) {
+            $method = 'TigerVNC';
+        }
+        _setSshProxy($method);
+        if ($proxy_jump) {
```
gfrenoy: This should also happen for RDP (or any other method...).

Shouldn't we do this inside _setSshProxy ?

hanspr: Yes, has been moved and tested on my last commit

hanspr: I have been testing and I have this strange behaviour, where the exit signal has been sent to the ssh tunnel, but in some occasions is ignored and the process will not actually close. The trace shows that the signal is been sent and the socket closed but the process remains in memory.

hanspr: Found the problem, we have to close the tunnel before any hard close instruction is executed. Actually the signal was not sent to the tunnel session because of this.

Created a generic function that is called before any hard close, so we can control any other things that would have to be handled before a hard close.

lib/pac_conn
hanspr: All this should be fixed.

lib/PACEdit.pm
```diff
@@ -225,6 +225,33 @@ sub __checkRBAuth {
...
+        _($self, 'vboxJumpCfg')->set_visible(1);
+        _($self, 'vboxCfgManualProxyConn')->set_visible(0);
+    } else {
+        if (_($self, 'rbUseProxyJump')->get_active()) {
+            _($self, 'rbUseProxyIfCFG')->set_active(1);
+        }
+        _($self, 'vboxJumpCfg')->set_visible(0);
+        _($self, 'rbUseProxyJump')->set_sensitive(0);
+        _($self, 'vboxJumpCfgOptions')->set_sensitive(0);
+        _($self, 'rbUseProxyJump')->set_label("SSH Tunnel port forward");
```
gfrenoy: To have the same wording ; shouldn't we label this "Use SSH tunnel" ?

hanspr: Done

gfrenoy:
I made some tests and it looks to work as expected, great ;-)

What do we do with the global Network Settings ?  Should we explicitly have this SSH tunnel option ?  Or is the same as Jump Server ?  For SSH, we will use the -J option and for VNC/RDP, we'll use SSH tunnel ?

hanspr:
>What do we do with the global Network Settings

I don't know. I could implement the same logic as in the local one, should not be very hard, I can do it if you like.

hanspr:
Finished changes, testing

gfrenoy:
I'm using this for some days days without any issue so let's merge it to loki and continue testing there.

Thanks again for your efforts and contributions Hans, we really made huge steps forward !

gfrenoy:
There is something wrong with some of my existing entries:
image

In the configuration file, for that entry, I see that it points to a directory:
image

The exact same entry without that PR was using the correct key:
image

gfrenoy:
Oh no, it's something I did not see before but that has been introduced with this change : d43e400774eb209a9eeee929da5c4c75ae64c657

hanspr:
That should happen only if you clic on the clear button, that I do not see in your image (strange)

That is the only way I found how to clear the box when clicking on the clear button

hanspr:
This is my image using the loki branch

imagen

I do not see the clear button in yours

gfrenoy:
> I do not see the clear button in yours

I cropped a bit too much my screenshot but the button is there, yes.

So when the value of ```public key``` in the configuration file is a directory, it does not load the value into the file widget...

gfrenoy:
Ok I think I understand where we are:

- In previous versions, it was possible to select a _directory_ as the ```IdentityFile```
- Ásbrú was silently accepting this and was gently adding it to the command line:
   (eg: ```ssh -i /home/my/directory xxx``` was called)
- In that case, SSH simply ignores and use the default IdentityFile (```~/.ssh/id_dsa```, ```~/.ssh/id_ecdsa```, ```~/.ssh/id_ed25519``` or ```~/.ssh/id_rsa```, according to the man page)
- In one of the recent changes, we now verify that the given ```IdentityFile``` is actually a file (hence not a directory)
- In that case, we just clear the value and raise an error when saving because no private key as been given.

Ok, now what I would propose:

- Keep the checks (eg: clear the value if it's a directory, it does not make sense, anyway)
- Accept an empty value for ```public key```, SSH will use the default if not specified ; this is cool :)
- Rename (None) to (Default) to make it clear that if you don't specify the private key in your connection, it will use the "system wide" default

gfrenoy:
Mmm, as far as I see in the [Gtk code](https://github.com/GNOME/gtk/blob/master/gtk/gtkfilechooserbutton.c#L2515), the (None) is hardcoded and there's no public method to change it.

Not sure either if there's a way to access the "private" button in a different way :-/

Maybe we can just put it in the tooltip for now that if (None) is selected ; the default system private key will be used ?  (it works fine, actually...)

PR #398, opened 2020-02-03
Find bug266

hanspr:
Fix new bug with Find in terminal, reported on the comments of closed issue #266


gfrenoy:
Should we apply this to master as well ?

PR #404, opened 2020-02-05
Vncrdpsocksproxy

hanspr:
I think this correction has to be made to support both methods, SOCKS or SSH Tunnel

I have tested my connection with SSH and still work (with this change applied)

Have no way to test your implementation.


gfrenoy:
Oh yes!  Wrong copy/paste, thanks for spotting and fixing the issue :)

PR #406, opened 2020-02-05
Vnc + rdp + ssh + socks 1

hanspr:
Added logic and configuration options to handle:

VNC/RDP --> SOCKS --> Tunnel Gateway --> Destination

Case

`ssh -o ProxyCommand='nc -x $proxy_ip:$proxy_port %h %p' -F $CFG_DIR/tmp/$uuid.conf -S $CFG_DIR/tmp/$uuid.ctl -f -N -T -M -L $LPORT:$IP:$PORT jumpserver`

Adding the proxy command is an option for VNC/RDP only and is enabled when using Tunnel.

imagen

It is preferred this way to avoid more complicated ifs in pac_conn

Still pending how to use user password, need a way to test and experiment.


gfrenoy:
I see the idea; that's an interesting case but such more complex cases can be fixed by creating several connections so I'm not sure we have to handle all types of combinations.

Indeed, in term of UI, I'm not sure this is very clear : we select one radio box and we use the options of another section.  Moreover, what if I want to use the global proxy settings instead ?

hanspr:
> what if I want to use the global proxy settings instead

Have not implemented, I guess that if we detect that the global proxy is set, then by default should use it, when you place it as global, it means use all the time.

If is not necessary we drop it, I really do not know how far you two want to go, with the current implementation I have no limitations.

hanspr:
> in term of UI, I'm not sure this is very clear 

We can move jump proxy up, and it will look better, and if we hide the radio button would even look clearer

hanspr:
But if we are not going to implement more combinations, that we drop it and try to finish the last one with authentication.

gfrenoy:
> But if we are not going to implement more combinations, that we drop it and try to finish the last one with authentication.

Yes we should probably put our imagination on hold for a moment, finalize the simple cases, write down what works, release something that is stable, wait for feedbacks...

We can also have another iterations later one.  With what is there, we can already do a lot :)

hanspr:
Fine by me.

PR #438, opened 2020-03-03
Fix437

### LINKED ISSUE #437: Loki branch - Auto saving feature
Hi Team ...

I do have the "Auto save config changes" unchecked in the Main Preferences.
image

Upon Asbru startup, the push button next to the Preferences button, has a grayed out "Save" legend:
image
 
Whenever I modify and save something in the Main Preferences dialog or in some Node configuration, the button gets updated with a grayed out "Auto saving ACTIVE" legend:
image

Why is the Auto Saving (apparently!) being turned on when I've it off in the Main Preferences ?
May you please explain how this feature is supposed to work ?

Thanks.

hanspr:
Fix #437 

This fixes the error on the button state.


gfrenoy:
Can you make the change to loki instead of master ?

gfrenoy:
Sorry guys but I don't believe this is the right fix ...

"auto save" --> will immediately save any configuration change
"save on exit" --> will save configuration before exiting

The button "Auto saving ACTIVE" should be shown when you have "auto save" activated.

And as discussed with popxunga, we do believe we should disable/hide "save on exit" when "auto save" is enabled.

Issue #441, opened 2020-03-04
asbru-cm freeze


skuran:
**Describe the bug**
asbru-cm v6.1.0rc1 main window hangs when a connection is made.

**To Reproduce**
1. Run asbru make a connection, just after the connection is made , main window freeze , the connections tab gets kindda grey and the mouse pointer turns into "busy" . The connection is also disconnects.

**Expected behavior**
No freeze expected.

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Version: 16.04

**Additional context**

When run from terminal:

$ asbru-cm 
INFO: Ásbrú Connection Manager 6.1.0rc1 (asbru-cm) started with PID 10015
INFO: Desktop environment detected : other
INFO: Using config directory '/home/meter/.config/pac'
INFO: Used config file '/home/meter/.config/pac/pac.nfreeze'
INFO: Using Gnome tray icon
*** unhandled exception in callback:
***   Can't locate object method "set_bold_is_bright" via package "Vte::Terminal" at /opt/asbru/lib/PACTerminal.pm line 456.
***  ignoring at /usr/lib/x86_64-linux-gnu/perl5/5.22/Glib/Object/Introspection.pm line 67.

skuran:
When I downgrade to 6.0.4 , freeze gone. 
$  dpkg -l | grep "libvte"
ii  libvte-2.91-0:amd64                        0.42.5-1ubuntu1                                          amd64        Terminal emulator widget for GTK+ 3.0 - runtime files
ii  libvte-2.91-common                         0.42.5-1ubuntu1                                          all          Terminal emulator widget for GTK+ 3.0 - common files

The-Course:
Let's discuss this in #435 and #434.

gfrenoy:
@skuran Commit https://github.com/asbru-cm/asbru-cm/commit/4004637770904cc9989bbee727e32fd60bb7ff4b should have fixed this.  It will be available in our next release.

In the meantime, you can try our snapshot release (see [instructions](https://github.com/asbru-cm/asbru-cm/wiki/installation-on-ubuntu)).

skuran:
Latest snapshot 6.1.0rc2 fixed 

Great !

Issue #435, opened 2020-03-03
6.1.0rc1-1 - When I open a saved session the application freeze


The-Course:
**Describe the bug**
When I open a saved session, the application crashes and I cannot select other sessions or the newly opened shell

**To Reproduce**
ASBRU version 6.1.0rc1-1

1. I open a saved connection/local shell
2. The application seems to be loading and is no longer interactable.

**Screenshots**

![image](https://user-images.githubusercontent.com/46918952/75759789-6a39a000-5d36-11ea-8734-650c7844c310.png)

**Desktop (please complete the following information):**
 - OS: Kubuntu
 - Version 19.10

**Additional context**
Running CLI asbru-cm I find the following output when I try to open a connection :
```
*** unhandled exception in callback:
***   Could not locate symbol vte_terminal_set_bold_is_bright at /opt/asbru/lib/PACTerminal.pm line 456.
***  ignoring at /usr/share/perl5/Gtk3.pm line 572.
```

gfrenoy:
This is also related to a wrong version of libvte.   ```set_bold_is_bright``` is available since 0.52 so I would think that you are using an older version...Let's discuss this in #434.

gfrenoy:
On 19.10, you should not have this message ... so the root reason is #434.  Let me close this for now.

PR #443, opened 2020-03-04
Fix434 5 41

### LINKED ISSUE #434: 6.0.4-1 - The tabs are not closed
**Describe the bug**
When I try to close a TAB it doesn't close

**To Reproduce**
ASBRU version 6.0.4-1

1. I open a saved connection
2. I close the tab -> I confirm that I want to close it
3. The TAB does not close

**Expected behavior**
That the TAB would close

 - OS: Kubuntu
 - Version 19.10

**Additional context**
Running CLI asbru-cm I find the following output when I try to close the TAB :

```
*** Vte::Terminal::feed_child: passed too many parameters (expected 2, got 3); ignoring excess at /opt/asbru/lib/PACUtils.pm line 3881.
*** unhandled exception in callback:
***   need an array ref to convert to GArray at /opt/asbru/lib/PACUtils.pm line 3881.
***  ignoring at /usr/share/perl5/Gtk3.pm line 572.
```

hanspr:
Fix #434 #435 #441 

Proposal to test capabilities instead of chasing version numbers.

We test what the current installed terminal library can or not do, set flags and then based on the flags apply the correct algorithms.

Have not tested the changes in PACUtils.pm with newer Vte version, only have 52


gfrenoy:
Great idea !

It looks like we are showing more are more information at startup ... Can we just display the VTE version at startup and display capabilities only if a "verbose" mode is enabled ?

hanspr:
>Can we just display the VTE version at startup and display capabilities only if a "verbose" mode is enabled

I will add the version in a minute

And do not see a verbose mode. And honestly when people send reports, the must common useful thing that we receive as support is that initial list of messages.

It also helps me when I'm testing new stuff, I usually run from the console.

When the application starts from the menu that stuff goes nowhere, so is not that it bugs anyone (or it does?). And when you run it manually, to report a bug, a want to have as much information as possible on that first interaction.

But if you still think that is to much, confirm and I will have to add a --verbose flag and  hide the loop information.

gfrenoy:
Oh really, we don't have that --verbose flag yet ?  Well, maybe a good opportunity :)

In "normal" mode, I think it is "scary" to see a very long list of kabbalistic texts ; having a clear, concise list of very informative instructions will increase the trust.

If we have the VTE version number, we can deduct the capabilities so I would only show this by default.

hanspr:
Ok

>If we have the VTE version number, we can deduct the capabilities so I would only show this by default.

This proves you wrong

```
} elsif (Vte::get_major_version() >= 1 or Vte::get_minor_version() == 52) {
        # Some distros have a special patched version of v0.52 that still requires 2 parameters; some others only one
        # Not nice but let's ignore the warning for that special case
        # See https://bugs.launchpad.net/ubuntu/+source/ubuntu-release-upgrader/+bug/1780501
        eval {
            local $SIG{__WARN__} = sub {};
            $vte->feed_child($str, $b);
            1;
        } or do {
            $vte->feed_child(\@arr);
        };
```

gfrenoy:
> This proves you wrong

Eh eh, good point :-)

But in that very exceptional case, we can ask for an additional contribution of the one that is reporting the issue, don't you think ?

I mean, my reasoning is : let's make it short and informative for 99% of our cases and let's keep the "verbose" mode for the other cases.

Or said again differently: why would you bug 100 people with an information that is only useful for 1 of them...Let's ask to that single person to do an additional effort.

hanspr:
Done

PR #449, opened 2020-03-07
Docs 2

hanspr:
Man, documenting and merging documents is a lot harder than coding, not that I did not know, but today was a painful reminder.


KlaasT:
Looks great for me. Thanks very much.

Sure documenting is always painful however users will thank us for it.

PR #450, opened 2020-03-07
Design improvements

hanspr:
Some minor improvements in the preferences ui, some settings can leave buttons bar accessible.

Moved save on every change next to save on exit so they are better visually related, even if save on every change is not specific to an on exit event, it still relates.

imagen

Added new my connections icon

imagen


hanspr:
@gfrenoy 

Any comments on this one?

gfrenoy:
Well, as you said, "save every change" is not related to "At exit" so it is not perfect to put it there... Anyway.

Issue #455, opened 2020-03-09
Loki Branch - Last commit on Preferences help button 


popxunga:
Hi ...

Just tested the latest Loki branch build asbru-cm-6.2.0~1583742140-1.el7.noarch and the today's commit on the Preferences help button looks like this:

![image](https://user-images.githubusercontent.com/37672088/76211743-d3ba2280-61fe-11ea-90cc-d878c68e498f.png)

If I put the mouse over the square button, the color reverses (gets white) and a tooltip stating "Open Online Help" appears bellow the square.

Thanks,

gfrenoy:
Does not work for me either:
![image](https://user-images.githubusercontent.com/981546/76258677-5a4d1f00-6254-11ea-9190-6cc885c6d6f9.png)

wcasanova:
it works for me.
![imagen](https://user-images.githubusercontent.com/12401985/76259991-94f38f00-6235-11ea-81a8-db44679f671e.png)
loki b25c417

hanspr:
Add a help icon in the resources in my last PR, to make sure it works across different distributions.

Must be some missing icon in your current icon set selected in your environments. Or bad icons indexes. So I guess there is no guarantee that using the default icons set will work in all environments.

KlaasT:
A new snapshot is currently building. Please have a look if it works for you.

hanspr:
mmhhh, I thing we already saw this in ubuntu, where there is this restriction where images are not allowed in button, only text.

Will add a button text on next PR together with a regression on the glade file

popxunga:
> mmhhh, I think we already saw this in ubuntu, where there is this restriction where images are not allowed in buttons, only text.
> 
> Will add a button text on next PR together with a regression on the glade file

Hi @hanspr ... Was the next PR merged or not yet ?

hanspr:
Nop, had to move it to PR #462 , began to have to many PRs around similar things pending to be merged.

popxunga:
Hi @hanspr 
With the last loki build asbru-cm-6.2.0~1584111928-1.el7, looks good and works:
![image](https://user-images.githubusercontent.com/37672088/76651725-ff217200-655c-11ea-931a-da317a634293.png)
Thanks.

hanspr:
Good, to know. Can not do any more with this ubuntu limitation.

popxunga:
Hi @hanspr ... Should we close this issue now ?

gfrenoy:
[...]

PR #459, opened 2020-03-10
Regression networking

hanspr:
Some regression happened in the glade file that radio options were not related any more in the networking tab.


hanspr:
Fix #455 

Fixed regression in glade file, and label help buttons, icons will not show for buttons in ubuntu by design.

hanspr:
Included in new link-help-4

Issue #463, opened 2020-03-12
Jump Host Feature: unable to estabilish connection, Missing user name in temp ssh configuration; 


danielemay:
**Describe the bug**
I Created a new connection using the Jump Host Feature, using private key for jump host and user/pwd for final host;
When connecting an error appear blocking the connection;
the final host username isn't forwarded in the temporary ssh configuration, blocking the connection attempt.

**Expected behavior**
Connection should be enstabilished.

**Screenshots**
![Screenshot_20200312_101851](https://user-images.githubusercontent.com/32195311/76507299-da3dd980-644c-11ea-94c6-116ec0add418.png)
![Screenshot_20200312_101908](https://user-images.githubusercontent.com/32195311/76507497-3274db80-644d-11ea-9c27-39be00162218.png)
![Screenshot_20200312_101923](https://user-images.githubusercontent.com/32195311/76507503-36086280-644d-11ea-9071-4c6dd4e43c5d.png)
![Screenshot_20200312_102030](https://user-images.githubusercontent.com/32195311/76507510-37d22600-644d-11ea-9b4f-e240ad6d1545.png)

**Desktop (please complete the following information):**
Linux 20190458-AIT 5.4.23-1-MANJARO #1 SMP PREEMPT Fri Feb 28 22:45:29 UTC 2020 x86_64 GNU/Linux

**Additional context**
Using Asbru Loki Version

hanspr:
The problem is in the user of the connection.

![imagen](https://user-images.githubusercontent.com/1572396/76535312-482fd400-6440-11ea-919e-2bb853abe1c4.png)

And by the litle information I can see in the red square, you must be using keepass. And the must current nightly build

If using KeyPass, then your current KeyPass settings are no longer compatible with the new implementation.

Your options right now:

1. Read the merge of this changes at #349 , to see how it works now.
   1. Basically, clic on the keypass button icon, and locate the entry that holds the current username and password for that connection.
    1. I'm in the process of documenting everything and this manual page is the next one, hope I'll have it by tomorrow, and then depends on the merging of documentation.
1. Write the user name and passwords directly in the fields, while you learn and reconfigure the new keepass implementation.

__Note__ : You must have installed keepassxc-cli ("KeePassXC") for this implementation to work.

hanspr:
After PR #464 is merged, an hour later you may check the documentation of the new keepass integration to see how it works.

https://docs.asbru-cm.net/

[...]

PR #464, opened 2020-03-12
added new keypass integration help docs

hanspr:
Added documentation for new KeePass Integration

Help for : #463


hanspr:
@KlaasT 

Hi, I think this property (in mkdocs.yml) is wrong, but I do not want to change it in case it affects the build.

imagen

I think that propertie should be:

site_url: https://docs.asbru-cm.net/

KlaasT:
Good finding. Thanks. Seemed to have missed a change there.

Issue #473, opened 2020-03-14
Loki branch - Change Expired Password


popxunga:
Hi Team ...

While changing an expired password I'm getting a dialog pop-up like this:
![image](https://user-images.githubusercontent.com/37672088/76681472-03946c00-65eb-11ea-838f-5e030421f2bd.png)

I enter the old password and then I get a new pop-up asking for the new password:
![image](https://user-images.githubusercontent.com/37672088/76681486-39395500-65eb-11ea-8468-49c3070f7f17.png)

Then I would expect to get a third pop-up asking for the new password confirmation, but I get nothing.
And when I try to paste the new password in the ssh window, the paste option is disabled in the context menu:
![image](https://user-images.githubusercontent.com/37672088/76681538-f75cde80-65eb-11ea-89b3-c35d0ff9737c.png)

Is there a way to avoid those GTK pop-ups ?
Or can we get the 3rd pop-up in order to complete the change password process ?

Thanks for your support.

gfrenoy:
This is a follow up of #138 and, indeed, it was not foreseen to check for password confirmation...

Should be possible though ; do you know how the remote server should be configured to ask for this confirmation ?

popxunga:
Hi @gfrenoy & @hanspr 

Not sure if I understood your question about configuration.

An expired password change process works like follows in any Unix system I worked with so far (except for root accounts where step 3 doesn't occur).
1- You login with the old password;
2- You get a reply stating that the password is expired and ...;
3- ... the old password is requested once again;
4- The new password is requested;
5- The new password confirmation is requested.

The existing process works fine from 1 to 4 and stops. 
But we are missing step 5.

Thanks.

gfrenoy:
> Not sure if I understood your question about configuration

To be able to reproduce what you see, I was wondering how to configure the remote SSH server so that is asks for a confirmation.  As far as I remember, when I did the change for #138, it did not ask me for a confirmation...?

popxunga:
I think there's no special configuration on SSH server's to achieve the described behavior, which I'm sure is the default on most operating systems and tested on Centos 7/8, Mint 19.3, Ubuntu Server 18.3, AIX, etc,. Here you have an example taken an Ubuntu Server:
![image](https://user-images.githubusercontent.com/37672088/76703005-1af45780-66c6-11ea-85de-5db4bfd56d30.png)

At least a setting to disable those ridiculous pop-up's should be provided to let us deal with the change password process directly in the terminal.

[...]

PR #476, opened 2020-03-15
Test expect

hanspr:
Proposed fixes for #473 

Several fixes on the login process and a new proposal to eliminate popup window for password prompts in ssh

* Encode regex prompts so utf8 characters can be used in regex.
    * The current implementation has tu use a trick like `[Cc]ontrase.a` to process `Contraseña`, but does not work porperly for the utf8 encoding comming from the terminal.
    * I would suggest to test and rewrite the password prompt shipped with asbru to: `(?mi)(pass(word|phrase)|contraseña).*?:\s*$` (after the full fix is accepted)
    * It handles grammar in other languages better (the default one does not work properly in spanish)
    * Or we could just document on the manual, alternative possible prompts 
* Fixed a wrong `if` that generates an extra password error if the password is left blank in the connection configuration.
* Removed popup for passwords and handles the prompt in the terminal, so it looks more natural and avoids the Gtk warning massages.

Peek 15-03-2020 12-04

* I know that we have a lot of options, but if we do not want to make it a fixed change that causes confusions, add option to choose popup or type on terminal prompt could be added.


gfrenoy:
Can you check [this commit](https://github.com/asbru-cm/asbru-cm/commit/84ea96dc68634d2eb6f4db9242127c60d59d0ae0) ?  I added on purpose what you are now removing ... It was related to [this discussion](https://github.com/asbru-cm/asbru-cm/issues/204)...

hanspr:
You right, reverted again.

I still get confused about the logic that to force password you need to type `<<ASK_PASS>>` for force asking for a password. I need to document this later in the documentation.

PR #481, opened 2020-03-17
Hide pass gtk warning

hanspr:
When calling the popup window to enter a value inside pac_conn, we have no reference of the container window, we are in a forked process.

This causes a warning, and that the dialog box shows behind the current window

imagen

The longest and more complicated solution would be to communicate with the Terminal object using the ctrl() subroutine so it opens the window for us, collects the value, and then we read it from the pipe.

I think that road is complicated and has the potential of data corruption if wrong encodings are applied.

The proposal is to hide STDERR, for the transaction.
Remove focus
Open the window
Recover STDERR
Return focus to the terminal.

And we get a clean interface and transaction.

imagen

The hideSTDERR and showSTDERR functions could be used in any other point where is needed if we face another similar issue that can not be fixed and represents just a warning.


gfrenoy:
Do we really want to hide without fixing the root issue ?

I mean: if we hide it ; it can have side-effects later that we won't understand because we don't see the warning, isn't it ?

hanspr:
ok

PR #485, opened 2020-03-19
Reimplement subst

### LINKED ISSUE #412: add <PORT> to PAC internal variables
We have USER.. PASS.. IP but..
I would like to have also PORT variable to use with local commands on non standard configurations.
eg. scp specific file to non standard port with one click

hanspr:
This has bugged me for many months now.

+ Recoded pac_conn `sub subst()` to reuse most of `_subst()` routine in `PACUtils.pm`
+ Reimplemented local variables substitution to work in a loop, easier to add more local variables later.
+ For a safe first round I differentiate the call from pac_conn to:
    - I excluded the substitutions that were not common to both implementations.
    - Preserve the original logic in pac_conn.
+ The changes have no logic change in any call to the original version `_subst()`

It has PR #484 (Fix #412) and PR #481 (hide gtk errors) already merged

Ran tests of following substitutions:

+ Local Variables
+ Global Variables
+ Ask
+ KeyPass 

## Configurations

+ IP = Global variable
+ Password = Title

imagen

+ We ask for a directory and cd to it
+ Echo local variables

imagen

## Result

+ Pulled password from title and logged in
+ Cd to dir after being prompted with popup window and no gtk errors
+ Printed local variables, including port from Fix #412 

imagen


hanspr:
same as previous one

gfrenoy:
Refactoring of "subst" looked like a great idea to me !

I did not merge it yet because I was waiting for the other one but it is not the same though ?  Do I see something wrong ?

hanspr:
It has the hide gtk error merged into it, so I need to redo the the gtk hide error end merge it to this again.

PR #487, opened 2020-03-19
Pipe pass gtk warning

hanspr:
Ok, this was interesting.

Removed warnings, correct positioning of window, had to go the long way using the socket with bidirectional communication.

Tests:

+ Ask password
+ Other enters in expect
+ Window now shows centred on terminal

imagen

During tests I noticed that and empty password in for User / Password option, would create a wrong login attempt. So the password `if` test to evaluate if we should prompt or not for a password was again rewritten.

Every time the expect finds the request for a password (passphrase) or we have the <<ASK_PASS>>  value we must prompt for a password, is not an alternative. Password less keys "do not generate a passphrase request"

From

```
if ((defined $PASS && $PASS eq '<<ASK_PASS>>') || ((!defined $PASS || $PASS eq '') && $AUTH eq 'publickey')) {
```

To
```
if ((!$PASS) || ($PASS eq '<<ASK_PASS>>')) {
```

Fixed along, some minor warnings, bugs in PACMain, and code from `_wEnterValue` that got confused during so  many merges.


hanspr:
In reference to the latest comment on  #463 

That some users have jumps hosts with password keys. The change on this PR, will not be able to automate the login, because we do not have the jump password implemented and the logic to know which of the 2 passwords is being requested at the moment.

But, with this PR, you are correctly prompted for the password and can login. After a failed attempt, but it works.

```
if ((!$PASS) || ($PASS eq '<<ASK_PASS>>')) {
```

So I think this is the right validation for all cases, so far.

gfrenoy:
Great improvement :)  This makes things way more clean !  Thanks a lot.

While testing I noticed KeePass was not working anymore ; I made some changes, let me know if this looks good ...

In PACMain.pm, is the change of " if (!$RUNNING{$uuid_tmp}{terminal}{_CLUSTER}) {" related to Gtk Warning ?

hanspr:
> In PACMain.pm, is the change of " if (!$RUNNING{$uuid_tmp}{terminal}{_CLUSTER}) {" related to Gtk Warning ?

Yes it was a comment in a thread from the password regex problem. And apparently sometimes might be undefined. So this way you test for both cases.

Checked the fixes, they are correct.

Issue #414, opened 2020-02-15
Support for non system wide installed instance of KeePassXC


gfrenoy:
**Is your feature request related to a problem? Please describe.**
If ```keepassxc-cli``` is not found as a system wide executable, KeePass integration is disabled.

![image](https://user-images.githubusercontent.com/981546/74595789-2eb88980-5046-11ea-86ab-3cda3961c074.png)

Hence, one might want to use the AppImage version of KeePassXC and in that case, the KeePassXC CLI is available through the following command:

```
./KeePassXC-*.AppImage cli
```

(see https://keepassxc.org/docs/#faq-appsnap-appimage-cli)

**Describe the solution you'd like**
Have a way to select the location of a specific version of ```keepassxc-cli``` or ```keepassxc``` (with "cli" to be added).

**Describe alternatives you've considered**
One can create a file ```keepassxc-cli``` in ```/usr/local/bin``` that contains:

```
#!/bin/bash
/path/to/KeePassXC-*.AppImage cli "$@"
```

hanspr:
This should be closed

PR #495, opened 2020-03-21
Fix414

hanspr:
Added binary selection, handle multiple changes, test for every new binary selected that it can work

added notification of error to inform password errors during password request.

update version, location and capabilities on every new binary selected

imagen

imagen


gfrenoy:
1) Tried to re-arrange a bit the UI to make it easier to understand (both from the code and from a user point of view).

2) I removed using ```\n\n``` to add space but using ```set_margin```

3) I also tried to support for the GUI version of KeePassXC ; in my case, I actually download the AppImage ; it looks to work but is very slow.  I think we are calling ```testCapabilties``` too often.
image

4) Question: what is the purpose of having "show protected yes/no" ?  Does it make a any difference ?

hanspr:
> what is the purpose of having "show protected yes/no" ? Does it make a any difference ?

It makes a difference in the flags that are set to call the `cli`. Internally it does matter is important not to remove the flags.

As a message? could be useful information when someone reports a bug and sends this image.
So I do not have to figure out what could it or not be set in the users current settings.

To the user? maybe makes no difference.

Is info that I was processing at the time when was figuring out how different versions were working, what did I had to do with each one, and I did not care to remove latter.

hanspr:
> it looks to work but is very slow. I think we are calling testCapabilties too often.

In my computer works really quick.

Maybe is your encryption level, or the keepassxc-cli is more efficient, I see that you are using a different binary. Test with `keepassxc-cli` to see if there is any difference and have a better understanding of this.

If there is a difference in performance, that we should warn and advice the user to use `keepassxc-cli` instead.

We need to test capabilities on every change to update the ui correctly, if do it at the end, it could escape the windows without a working environment. I preferably like Apps that warn me immediately and not at the end when I though everything was OK.

hanspr:
Stand by with this one, noticed a bug was introduced with one of your modifications, will push fix in a few minutes

gfrenoy:
> > what is the purpose of having "show protected yes/no" ?

> It makes a difference in the flags that are set to call the cli.

Oh OK I got it now : with the newest version of ```keepassxc-cli```, you have to use ```---show-protected``` to see values that you would see before in clear text.  So it is required to know if this flag is supported (or not).

Thanks for the clarifications.

gfrenoy:
> > it looks to work but is very slow. 

> In my computer works really quick.

Mmm, "very slow" is indeed a bit exaggerated :)  It is reasonably quick.

> I think we are calling ```testCapabilties``` too often.

This is my point actually : we should only call ```testCapabilties``` when necessary. At least with my code, we call it twice if you don't specify the "cli" version.  Maybe a warning would be a good idea, indeed :)

hanspr:
Ok, on every startup we need to make sure keepass is working and if it has been updated since the last boot, know its working, and that what can it do.

We need 2 calls as a minimum, one for the version number, one for the capabilities.

The extra one (is if the user messed up choosing an invalid binary), as long as the binary is correct there will be 2 calls at the beginning.

And an additional call every time the users changes the binary, only then.

If you want to only call at the end, fine, It does not bug's me to do it on every selection change.

With this change the extra (3rd or 4th call generated by testing `$CLI cli`, should not be necessary. I know that is not necessary for keepassxc-cli binary, only for the AppImage binary.

gfrenoy:
https://github.com/asbru-cm/asbru-cm/pull/495/commits/4541ba70d50c1491feea1c8932f105f08ba99ebe  much better :)

Still a doubt about this test: ```$$self{kpxc_pathcli} =~ /appimage/i```

It means the user would not have renamed the file or that the user did not select another "AppImage" not related to KeePassXC...

hanspr:
> It means the user would not have renamed the file or that the user did not select another "AppImage" not related to KeePassXC...

There is a point where this has to stop. If they want to make it work, we should align to known standards, the keepassxc standards. And we should program for those standards. And not chasing all possible user's oddities.

If you really want to give that flexibility, then we shouldn't have a file selection box, but a text field so he/she writes its full path and parameters. Or we add an extra text box for "additional parameters". In any case if the users enters wrong data will not work either.

hanspr:
Would it be better if we test like this?

```
        if (!$$self{kpxc_cli} && $$self{kpxc_pathcli} !~ /keepassxc-cli/i) {
            $$self{kpxc_cli} = ' cli';
        }
```
That would add cli to any other name different that the normal keepassxc-cli

But again if we have to cover for a user changing the binary name to : `unlock-my-database` then we are doomed.

gfrenoy:
> Would it be better if we test like this?

Mmm, likely, yes.

What I tried to do is to call the selected file and look at the output to figure out if it was the correct CLI version of KeePass, the GUI version or anything else.  My code was probably bad but the idea was really to validate the selected file was actually a KeePass binary ; whatever the name.

> then we are doomed

I all cases I think we are :-)  But thanks to you, it's a real pleasure !

Have a great evening/night.

hanspr:
Ideas for tomorrow.

```
file KeePassXC-2.5.3-x86_64.AppImage 
KeePassXC-2.5.3-x86_64.AppImage: setuid, setgid, sticky ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/l, for GNU/Linux 2.6.18, stripped

file /usr/bin/keepassxc-cli 
/usr/bin/keepassxc-cli: ELF 64-bit LSB shared object, x86-64, version 1 (GNU/Linux), dynamically linked, interpreter /lib64/l, for GNU/Linux 3.2.0, BuildID[sha1]=045e56a96f787b407b16363ac1c5fa8efa1a245c, stripped
```

Trying to figure out the type of binary could help to decide if we should use cli or not?.

AppImage gives `LSB Executable` , keepassxc-cli `LSB shared object`

https://docs.conan.io/en/latest/integrations/deployment/appimage.html

If it does, it will improve setting the parameter with better accuracy than just the name.

hanspr:
Ok, was not aware of the concept of an AppImage, now I know.

I did notice that start an AppImage is slower.

Considering this, it go really hard, I added:

+ Try to indirectly detect if it is an AppImage what the user selected to avoid the 3rd test
+ Removed the KeePassXC testing if integration is not enabled to avoid testing on start up when disabled
    - This now causes to test when enabling to check if we have it or not.
    - faster start up
+ The introduction of a selectable binary, introduced the need to control more possible configuration options.
    - No cli installed, but gives the opportunity to select one.
    - If user leaves incomplete information we must remove everything to avoid incomplete states
    - We force the UI to update again so to reflect the final result.
+ Lots of changes to control valid, invalid binary selected, database selected or not, keyfile available or not, etc. And update the UI and messages accordingly.
+ Manged to reduce the amount of tests during startup to only one.
+ Added verbose messages

hanspr:
Literally, added magic to this last push

https://github.com/AppImage/AppImageKit/issues/521#issuecomment-345434425

https://github.com/denysdovhan/inboxer/issues/7#issue-285250824

So we can detect that is an AppImage and add `cli`.

Of course we can not know that is a valid AppImage, but it will fail on he version number must likely.

hanspr:
Ok, I think that I have finally nailed this one down.

It got a lot more complicated than I thought :tired_face:

hanspr:
Sorry, I got so excited about detecting the AppImage, and removing multiple calls  to testCapabilities, that I broke the code used to pull the passwords during the connection.

Is fixed.

I renamed `_testCapabilities` to `_setCapabilities` to make it clear that is not just a test, but actually setting the environment to correctly call the cli, in this case during connections (pac_conn) is important to set up what can or not do the current cli.

Test ran:

+ Selecting an old version of the cli : 2.4.3
+ Open an SSH connection with keepassxc masks
+ Let the session opened
+ Change the setting of KeePass to use the newer version : 2.5.3
+ Open a second SSH connection with keepass masks
+ It opened correctly boths sessions even when changing or updating the cli in between

hanspr:
I think this one is finished, unless you have some concerns.

How busy are you today to see if we can check, fix (if needed) and finish this one.

gfrenoy:
Yes, would be great to finalize this ; I think we are very close !  I noted the following:

1) Should we really show that scary message ?
image

- Not sure this is correct
- If I use the workaround described in #414, I'm still using the AppImage but don't see the message
- what is actually slower is to detect that we don't use the CLI binary but the GUI binary (hence we need add "cli" ourselves)

--> maybe we can write down in the documentation that pointing to the CLI binary is recommended

2) If I select something stupid (hence, not a valid keepassxc binary), I would show an error message.  He, we just fallback to the default without drawing the attention of the user that she/he did something wrong.

3) About the "debug" messages in verbose mode, shouldn't have the same prefix everywhere ?
  - INFO:
  - WARN:
  - or DEBUG:
That might make things easier to filter later on.
If we want to be able to filter by module, we can add a second prefix (DEBUG:KEEPASS: as example) ?
If we keep this standardize, it might be easier later to have global helpers.

hanspr:
1. I'll remove it makes no difference to me

2. It does fall back to default one, without warning. Will add the warning

3. Ok, in a minute

hanspr:
done, next round

hanspr:
Should I merge PR #527 into this and remove the other PR?

gfrenoy:
Brillant!  Thanks a lot :)

PR #497, opened 2020-03-22
Fix182

hanspr:
+ Save previous config dir as dir.old
+ Moved directory
+ Rename config files
+ Fix paths in yml
+ Migrate user paths too
+ Renamed icons
+ Renamed pac_conn to asbru_conn

You may finish it with additional messages or other considerations not included.


hanspr:
Maybe I over did it?

Should I return `asbru_conn` to `pac_conn` ? to avoid the tests failures?

Or you want to continue with this moves?

hanspr:
Forgot to mention, but this has been tested, I'm currently running asbru with all this migration applied and the process was completely transparent.

It also migrates any previous setups you might have created with the `--config-dir` option.

gfrenoy:
> Should I return asbru_conn to pac_conn ? to avoid the tests failures?
> Or you want to continue with this moves?

No big deal, you just need to replace ```pac_conn``` with ```asbru_conn``` in the RPM spec file :-1:  ```dist/rpm/asbru.spec``` and this should pass the test.

> Forgot to mention, but this has been tested, I'm currently running asbru with all this migration applied and the process was completely transparent

Looks great.  Not much time to look in detail but I'll definitely will soon !

PR #503, opened 2020-03-24
Fix499 2

### LINKED ISSUE #499: text color in dark plasma themes
**Describe the bug**
friends in the data entry boxes, the cursor is not appreciated, I do not know if it is a problem with the plasma gtk theme "breeze-dark", the problem is that the text is not well appreciated, or the position of the cursor in the text. I present this problem from previous versions

**Screenshots**
imagen

 - OS: KDE/Plasma , Archilinux
 -  Asbrú 8f6e217

hanspr:
Fix #499 

Added dark theme: intended for users with system dark themes

Added system theme: it honours the system theme

Removed caret-color


gfrenoy:
Hans, will it be OK to maintain all those themes in the longer run ?

I mean: the more themes we add, the more work it will be each time we want to add one additional icon, isn't it ?

hanspr:
I do not know how to go around the dark theme issue if we do not have one explicit for it

The announcement of ubuntu not supporting the caret-color property, made that impossible.

I can remove the system one, if you like. Or just have : asbru-color and asbru-dark

Eventually I'm thinking to allow to install themes in the config directory, but that will be ahead of this.

Tell me what you want, I do not know in this case what should be the best option.

gfrenoy:
I understand why you did it (and this is great ;-p)

I'm just wondering how many themes we would like to maintain inside the "main" project ?  And what are the themes that need to be in another projects / maintain separately...?

hanspr:
Don't know.

My reasoning is:

+ One dark for dark users
+ One light for light users
+ One system, for all other things we are not considering.

Now the only one that could be removed, and then we have to fix the code is the Default one with gray icons.

I think, that as soon as we open the option to install themes in the config dir, eventually no one will care about the ones we publish (hope someone will come up with something way cooler than what we have right now) and we will maintain them to the minimum.

hanspr:
Forgot to mention, but the theme "system" has absolutely no fancy hover effects , etc to avoid conflict with all this system themes.

So, if you really want to take it to the minimum, we just realease a system set, and then loose what we have about hover effects, etc.

gfrenoy:
So, yes, 3 themes is the very minimum.  We can probably afford a fourth one but I guess we don't want to add more and indeed leave it to some creative ones :)

Let's try like that.

Thanks a again.

PR #506, opened 2020-03-25
Compact close to tray

hanspr:
In Compact mode, the close to tray icon should be set always.
On a new install if the user changes to Compact mode the flag is hidden but the default value of false breaks the functionality


hanspr:
Included in PR #513

hanspr:
Should have closed.

PR #513, opened 2020-03-26
Fix294

### LINKED ISSUE #294: Asbru entry left on Window List (el7+el8)
Hi ...
Seeing this on asbru-cm-6.0.0\~1575196208-1.el7 and asbru-cm-6.0.0\~1574600188-1.el8.
The "Hide to tray instead of closing" is not working correctly.
Once the asbru-cm main window close button is pressed, the window get iconified to the systray, but an entry remains in the Window List bar at the bottom (the entry should be removed as in the previous versions).
Note that on CentOS 8, the Window List extension has to be activated via gnome tweaks.
Thanks

hanspr:
Fix #294 

Tested

+ Ubuntu 19
    - With unite extension
    - No unite extension
   - __Note__ : have not found a way to know if the unite extension is enabled or not. Only can test if installed. So, if installed and disabled will create an inaccessible tray icon.
+ Centos 8
   - As per #294 
+ Mint


hanspr:
Need a few minutes with this one, I have to apply the same logic as in Ubuntu, find if a tray extension is installed and if not minimized, to avoid missing the icon on the tray.

hanspr:
Improved logic, the rule would be

+ If I can confirm there is a known extension, then enable close to tray
+ If I can not confirm. Then minimize. The worst thing it can happen is that someone reports a similar issue again. And we will have to add more extensions that allow tray capabilities. But there will be no cases of inaccessible tray icon.

this should be ready

PR #515, opened 2020-03-26
Fix182

hanspr:
Did it again deleted the wrong branch

+ Save previous config dir as dir.old
+ Moved directory
+ Rename config files
+ Fix paths in yml
+ Migrate user paths too
+ Renamed icons
+ Renamed pac_conn to asbru_conn

You may finish it with additional messages or other considerations not included.


hanspr:
Comments.

To test this modifications. I would recommend.

Copy your real data to a backup location.

If you have alternet configurations, you can test with them

First test with --config-dir=other

Very time you migrate, there will be a directory.old version.

remove the new one, rename the old one, and you will be back to the previous  configuration.

hanspr:
Added popup warning and message with options

You may edit the message to adjust to your liking.

hanspr:
@gfrenoy , @KlaasT 

To test:

Manually back up your pac directory

```
cp -Rfp ~/.config/pac ~/.config/pac.save

Check pac.save exists
```

Upgrade to new configuration directory

```
git checkout fix182

perl asbru-cm [--verbose]
```

Downgrade to previous version

```
git checkout fix-downgrade

perl asbru-cm
```

All branches should work

----

If we feel confident that we have a working upgrade version. Then the must secure path would be (as I understand it).

+ Release `downgrade-master`  as : __asbru-cm 6.1.2__
    - This has to be released before __asbru-cm 6.2__ so users from official master release can downgrade in case of an unexpected error.
+ Release `fix-downgrade` as : __asbru-cm 6.1rc3 or whatever__
    - This has to be released before __asbru-cm 6.1rc4 or whatever__ so users from loki can downgrade in case of an unexpected error.
+ Release `fix182` __asbru-cm 6.1rc4 or whatever__, to invite brave testers of the migration. And improve testing before release of the official final version.
+ Merge loki to master as : __asbru-cm version 6.2__

---

Summary:

+ Upgrade
    - Creates a copy of the current config directory : pac -> pac.old
    - Rename pac -> asbru
    - Rename files and other data in asbru.yml
+ Downgrades
    - Remove directory : `asbru`
        - So a user can rerun a new fixed version to upgrade
    - Rename : pac.old -> pac

hanspr:
When will we dear to take the jump?.

This is what is left to finish your milestones.

And I'm waiting for this to be able to push bunch of improvements on the application based already on this branch.

gfrenoy:
Will give it an extensive try this weekend.

hanspr:
Sorry for the last minute change.

I do not like the idea to leave the migration code permanently inside asbru-cm

So I moved it outside to the "utils" directory, so one day we can remove all the logic for a migration to clean up the code. And if anyone jumps from pac to asbru 7,8,? . They will have to run the script manually to convert the file.

it will be as easy to run 

`/opt/asbru/utils/pac2asbru.pl /home/user/.config/pac /home/user/.config/asbru`

hanspr:
@gfrenoy 

If you do not mind, I would like to freeze any more updates to loki, until we finish this and merge it, because the changes in this branch produce a lot of merge conflicts.

And if I will eventually merge something wrong and break what is already working.

gfrenoy:
> I would like to freeze any more updates to loki

Ouch, did not read all of this in the right order ; sorry.  I'll merge the latest change.

gfrenoy:
Ok so you fixed it already !  Great :)  Let's stop from changing anything in loki until we finalize this !

popxunga:
Hi ...
Probably not a very common scenario, but along with other important program configurations, I keep the pac config directory in another file system and I point to the pac config directory by creating a symlink  in ~/.config. Will the configuration upgrade process work in this case ?
Thanks.

gfrenoy:
>  I point to the pac config directory by creating a symlink in ~/.config. Will the configuration upgrade process work in this case ?

Mmm, interesting point !  Thanks.

I gave it a try and the upgrade will work but you will not have a backup copy !  Ouch, that's not the idea :-/  We'll need to think about this !

> Is the configuration upgrade process already merged into loki ?

Not yet.

The idea, the plan is to wait for a few more days for feedback (hence the announcement)...

gfrenoy:
> I point to the pac config directory by creating a symlink in ~/.config. Will the configuration upgrade process work in this case ?

Ok should work fine now ; the backup copy is using the option "-L" of [cp](https://linux.die.net/man/1/cp) to dereference any symbolic link and make a "real" copy.

Thanks again for the good feedback!

popxunga:
Thanks for that @gfrenoy :-)
Maybe for later consideration, but it would be nice as well to have a way to specify the base directory where to read/save the configuration.

gfrenoy:
The idea is:

- Either you let it happen "automagically"
- Either you want to crontol what is happening and you are using the dedicated script ```utils/asbru2pac.pl``` before starting the new version

Said differently: with ```utils/asbru2pac.pl``` you can specify the source and target directories of the migration.

popxunga:
Agreed. 
I was thinking a bit outside the spirit of this change. I mean that would be nice to have a way in the Preferences to specify where do we want to save/read the configuration, instead of having the configuration based in a fixed directory as it is currently in ~/.config.

hanspr:
@popxunga 

We need a fixed point to start, otherwise how do you find the directory you want if it is in the config file. And reading configuration files twice or trice to jump from one to another, to another should be unnecessary. 

I think that the options in the command line should suffice for this options. The config-dir option lets you set an absolute path.

https://docs.asbru-cm.net/Manual/CommandLine/

popxunga:
Hi @hanspr

You're correct but we could use something similar to what Mozilla does, I mean save a sort of configuration profile only in the ~/.config/asbru-cm folder. And that configuration profile would have the target folder for the user configuration.

I was not aware of the mentioned command line option. Thanks for that. It is more than enough to me 😁

Thanks!

gfrenoy:
So here is our plan:

- [X] merge support for downgrade from migrated files into master branch
- [ ] release 6.1.2 to included the latest fix and support for downgrade
- [X] announcement that a configuration change will happen in loki ; warn users to be careful the next time they upgrade
- [ ] merge 515 into loki
- [ ] code freeze, tests and feedback in loki 
- [ ] release 6.2.0~rc1 on master --> first release with migrated configuration file ; 6.1.2 become legacy with downgrade support
- [ ] at the same time, move loki to 6.3.0 --> for the next bunch of changes (to be decided...)

KlaasT:
How long would we do the test period in loki? 2 weeks?

PR #523, opened 2020-03-27
Fix warning unite

hanspr:
No tray available warning should be shown only if no tray detected.

Removed unite, because other distros may use a different solution, best option would be to document options based on distros in the documentation.


gfrenoy:
Yes, indeed.  Can we already foresee a link to the page in the documentation ?

hanspr:
Sent e PR with a basic page, when published the link will be:

https://docs.asbru-cm.net/Manual/Preferences/SytemTrayExtensions/

gfrenoy:
Thanks a lot :)

Just realized that the link in a "tooltip" is not really useful (you have to copy it "manually").  I just added some words to suggest reading the documentation for more information.

PR #531, opened 2020-03-29
downgrade config dir migration to homologate all the code

hanspr:
Before merge #515 

We should prepare master to be able to downgrade, and publish a version that can downgrade from the upcoming config migration.

### DIFF
--- a/asbru-cm
+++ b/asbru-cm
@@ -120,6 +120,12 @@ our $VERBOSE        = 0;
 # Load PACMain now, so it can read the global $CFG_DIR
 print "INFO: $APP_NAME ($Script) started with PID $$\n";
 
+foreach my $arg (@ARGV) {
+    if ($arg =~ /-{1,2}verbose/) {
+        $VERBOSE = 1;
+    }
+}
+
 # Check for Unity's systray-whitelist presence of 'asbru'
 my $wl = `gsettings get com.canonical.Unity.Panel systray-whitelist 2>/dev/null`;
 chomp $wl;
--- a/asbru-cm
+++ b/asbru-cm
@@ -104,7 +104,6 @@ use warnings;
 our $CFG_FILE_NAME  = "pac.yml";
 our $INIT_CFG_FILE  = "$RealBin/res/$CFG_FILE_NAME";
 our $CFG_DIR        = $ENV{"ASBRU_CFG"};
-our $CFG_DIR_OLD    = "";
 our $CFG_FILE       = "$CFG_DIR/$CFG_FILE_NAME";
 our $BACKUP_DIR     = "bak";
 our $APP_NAME       = PACUtils::_appName();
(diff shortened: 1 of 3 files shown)


gfrenoy:
Downgrade on 6.1.x (and earlier) will be a bit more complicated, isn't it ?  I mean: ```pac_conn``` does not exist any more in the migrated directory :-/

hanspr:
>pac_conn does not exist any more in the migrated directory :-/

I understand for "downgrading" remove the latest package and then install one older, the older one will have pac_conn in it and everything like it was. The only thing left is recover the old saved directory.

pac_conn is not saved in the config directory is part of the application in `/opt/...`

What did I miss?

gfrenoy:
This version should be very complete, the options are:

- Try to downgrade from ```~/.config/asbru```
- Try to rename ```~/.config/pac.old```
- Create a brand new (empty) configuration

At each step, one can chose to proceed or cancel.

If believe only the migration script ```asbru2pac.pl``` should be kept in 6.2.0 ; the migration code in the main ```asbru``` file is not useful when starting the new version ; do I see this right ?

hanspr:
Its a matter of how far you want old PAC users that have not come back  (or asbru pre 6.1 users) and find themselves with asbru version 6.3, 6.4, ... 10, and they want  to jump all the way to the latest release with their old config dir available.

So an alternative would be to move the code to the documentation section and explain how to download and run manually after is been removed from version 6.2. Or you support this for a full 1 or 2 years and not in term of versions.

gfrenoy:
Sorry, I don't see the scenario for which you'd need that portion of code in 6.2 (or later):

https://github.com/asbru-cm/asbru-cm/blob/37656efed7a1e2a8fb3e83e05669f42f5a25bd3a/asbru-cm#L145-L182

?

hanspr:
I'm sorry, I'm lost with this.

That code is not into fix182, I do not think was ever meant for 6.2, as I mentioned before, was for an intermediate package for master users.

I agree that for 6.2 the final version should be the one from fix182.

I guess that some kind of merging has to be done to accomplish to remove it later when you merge loki into master; and this conversation is about all that.

Sorry, my __git__ skills are a bit limited yet, so I do not know how to perform those moves.

But for the sake of learning, guess is something like.

* merge asbur-cm from master into loki
* pull loki
* merge loki into fix182 and fix merge conflicts of asbru-cm
* push new version of fix182

??

gfrenoy:
> I agree that for 6.2 the final version should be the one from fix182.

Ok good so I will close #532 

> I guess that some kind of merging has to be done to accomplish to remove it later when you merge loki into master; and this conversation is about all that.

No it's because #532 was pending and I did not see why we should do that:

- either we do not merge #515 --> not necessary for the mass (only for the few audacious ones who tried your branch ;-p)
- either we do merge #515 --> not necessary since once it's migrated, not need to downgrade in 6.2

So, I think we are OK with this.

Let me put the next steps in #515.

PR #532, opened 2020-03-29
Fix downgrade

hanspr:
Before merge #515 

We should prepare a loki version that can downgrade the upcoming config dir migration.


gfrenoy:
Shouldn't we also ask for confirmation before proceeding with a downgrade ?

Especially if we delete what is in the new ```asbru``` directory ; let's image the following:

- I start the new version --> ```pac``` is copied to ```asbru``` and I have a safe version ```pac.old```
- I play a bit with the new version, it works fine for a while ; I start making changes in ```asbru```
- For some reason, I start an old version ; ```pac.old``` is reverted to ```pac``` and ```asbru``` is deleted without any notice ; my changes are lost :-/

hanspr:
Downgrade warnings added.

gfrenoy:
Hans, sorry that I keep thinking and hesitating about this but I'm not confortable with this "destructive" approach :-/  I would really leave this to the user to do it outside Ásbrú.

I mean: even if we put a clear warning, I do fear they don't read it and loose all their data.

While playing around with this, I noticed I actually had another strategy to cope with both versions:

1) for the new version, I was using ```.config/asbru```
2) when starting the old version, I was actually using the ```--config-dir``` parameter to use the ```.config/pac.old``` directory

That works quite well since I just wanted to test something on the old version ; all my environment "looks like" the new configuration but without the latest changes.  And I know that if I make a change there, it will not be available for the new version.

So what would think about this:
- When detecting that ```pac``` does not exist but that ```pac.old``` exists, ask if they want to load ```pac.old``` or create a new clean ```pac```
- In the new version, if configuration in ```asbru``` does not work properly and that ```pac.old``` ; show a warning suggesting to backup then delete the directory are proceed with the migration again.

Maybe too complicated ?  But as this can have a major impact for many users, I think we should really think twice on this.

Also I was wondering if this is really required for the loki branch ?  I mean, if we merge 515 into loki, this code is not necessary ; do I get it right ?

hanspr:
Ok, this versions are none destructive upgrades and downgrades.

They will backup up to 10 new or old config versions.

On each downgrade or upgrade the current configs will be upgraded or downgraded. So you can 

+ upgrade, add, remove stuff
+ downgrade, add, remove stuff
+ upgrade, add, remove stuff
+ downgrade, add, remove stuff
+ upgrade, add, remove stuff
+ downgrade, add, remove stuff
...

And if you play around upgrading and downgrading you will fill up your old# and new# backups, so is up to you clean up the left overs if you moved to much up and down. And you will have enough saved configuration files to choose from in case of any recovery.

image

hanspr:
> Also I was wondering if this is really required for the loki branch ? I mean, if we merge 515 into loki, this code is not necessary ; do I get it right ?

It depends, if you merge this, then wait for a released loki package cloud update, that one will have the downgrade code. And after that you merge 515 and you will get a upgrade package. Now you will have a middle loki build to create a downgrade for loki testers. 

That is how I understand it.

In other words, if you do not merge and release a loki package, loki testers will have to downgrade manually in case of any problem. And we will not have a package for testers to test a downgrade.

KlaasT:
We are keeping a small History of packages inside the repository. So a downgrade should be possible. By selecting the version using apt for example

hanspr:
You 2 know better, so decide.

The important one I think is the one for the master branch because the users from that branch are the ones that most likely are not interested or want to spend time figuring out how to manually downgrade if they have to.

Again, as I understand it.

gfrenoy:
What I had in mind (but it's only a wild idea at this stage) for the next steps was:

- release  6.1.2 to included the latest fix and support for downgrade from migrated files
- announcement that a configuration change will happen in loki ; warn users to be careful the next time they upgrade
- merge 515 into loki
- code freeze, tests and feedback in loki
- release 6.2.0~rc1 on master --> first release with migrated configuration file ; 6.1.2 become legacy with downgrade support
- at the same time, move loki to 6.3.0 --> for the next bunch of changes (to be decided...)

For 6.0.4, code is frozen and there is no support for (automatic) downgrade,

Does that make sense ?

KlaasT:
That sounds doable. I suppose 6.2 will have then some breaking changes which probably need some time and polishing?

I agree that we shouldn't fiddle much with the 6.0 codebase. We don't have much telemetry on which versions are in use but I could imagine that 6.0 should be fading out quite well on the clients.

gfrenoy:
>  I suppose 6.2 will have then some breaking changes which probably need some time and polishing?

It includes many changes but many users are already using it and most of the feedback we are receiving today are related to 6.2 so it's already well tested and is reasonably stable.

KlaasT:
Ah okay. Just checking in case anything was missed. Then this should be a reasonable step. From the master snapshots and releases it is also possible to do a downgrade using apt.

hanspr:
> I suppose 6.2 will have then some breaking changes which probably need some time and polishing?

There are braking changes, they are related to branding. We are renaming many files from pac to asbru:

+ ~/.config/pac to ~/.config/asbru
+ pac.yml to asbru.yml
+ pacc_conn to asbru_conn
+ and inside the code images references, etc.

A wrong move will cause that user will have no access to his/her configurations. Nothing would be lost. But the user will be very confused looking to a blank list of connections.

But is only a one time "risk", if I may phrase it properly like that. Once the renaming of the config directory and files is done, there should be no further issues.

I've been running and using daily the new version for 12 days, and have not had any issues yet.

gfrenoy:
If you don't mind, I'll merge this into #515.

gfrenoy:
Let's close this one and only "cherry pick" the asbru2pac.pl to #515.

Thanks/

PR #546, opened 2020-03-31
Fix warning keepass

hanspr:
If there is no keepassxc-cli in the system and we open the settings window, on close you will get a warning even if keepass integration was not enabled. So it warns only if you set it and something was wrong at the end of the configuration.


gfrenoy:
Saw that one as well !  Thanks for the fix :)))

PR #547, opened 2020-03-31
Fix533 2

hanspr:
Dropping `get_background_color` is unreliable even if it exists.


hanspr:
@gfrenoy 

This last version has the defintive fix to #533 the issue was in a different place. Found by @popxunga.

Now, this also takes in consideration that Debian 10 crashes if we try to use get_background_color method.

PR #548, opened 2020-04-01
Docs explanation about transparency and side effects

hanspr:
Added explanation about the impact of turning transparency on.


gfrenoy:
Is this OK ?
image

I would think we also need to install the ```material``` theme first ?  I guess it can be done in [deploy_docs.sh](https://github.com/asbru-cm/asbru-cm/blob/master/scripts/deploy_docs.sh) but I'm not sure at all (and did not test this yet)

hanspr:
Removed material then

PR #568, opened 2020-04-10
Add settings to resize side tree panel automatically

syco:
Hi all,

I find really annoying that the side panel (with the connections tree) changes size when I click on it, or when I open a terminal.
I know it's only a few pixels back and forth, but I added a listener to resize it to a fixed width I like when that happens.
Please let me know if you think this could be implemented better.

Thanks

### DIFF
--- a/lib/PACConfig.pm
+++ b/lib/PACConfig.pm
@@ -221,6 +221,9 @@ sub _setupCallbacks {
     _($self, 'cbConnShowPass')->signal_connect('toggled' => sub {
         _($self, 'entryPassword')->set_visibility(_($self, 'cbConnShowPass')->get_active());
     });
+    _($self, 'cbCfgEnableForceTreeSize')->signal_connect('toggled' => sub {
+        _($self, 'cbCfgForceTreeSize')->set_sensitive(_($self, 'cbCfgEnableForceTreeSize')->get_active());
+    });
     _($self, 'cbCfgPreConnPingPort')->signal_connect('toggled' => sub {
         _($self, 'spCfgPingTimeout')->set_sensitive(_($self, 'cbCfgPreConnPingPort')->get_active());
     });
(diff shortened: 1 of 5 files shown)


gfrenoy:
Thanks a lot :)

Any chance to have the CLA signed ?

In any case, I'll probably try to merge this into loki instead of master ; we should be close to a new release, anyway.

syco:
I thought I did it.. anyway, it's signed now.
Thanks for considering the pull request.

hanspr:
@syco 

Would your change stop the current free resize of that pane to work? will it fix the size once you set the new option?, right now a user can resize that panel freely.

imagen

I have tested my version and I do not see any size jumping. So I think it would be important to understand the case when it happens and why it happens. It sound like some sort of environment issue with your linux distribution and your current theme and themes applied by asbru.css files.

pane_resize

Would you mind to open an issue request to check the case that you describe, and the steps to reproduce.

Please, include your asbru version number, linux distribution and version, try the must recent asbru version if possible, and share the important configuration settings that you have set. Include images or a gif movie to help us see the issue, it does help a lot.

If you are using asbru version 6.1.2+ or the current loky branch, it would be safe to export a debugin  yml file to help us reproduce the problem. 

https://docs.asbru-cm.net/Manual/Preferences/Preferences/#export-settings

syco:
Hi,
I use version 6.1.3 from the master branch, on kubuntu 19.10.
Here's the debug configuration ([debug.txt](https://github.com/asbru-cm/asbru-cm/files/4512661/debug.txt)),
but I think I found part of the issue, I sync the config folder with other computers, and because pac.dumper pac.nfreeze pac_stats.nfreeze pac.yml.gui and pac.yml.tree often create conflicts I only sync pac.yml,
so when I start asbru it doesn't remember it's position and size, and this happens:
ezgif com-video-to-gif

but sometimes it also happens when I close a connection or open more.
For now I'll remove the sync and see if it still happens.
Thanks..

hanspr:
Tanks, @syco 

This is interesting because all tests I've done, are with the panel on the left. Is in those little details where de devil always hides.

I can confirm it happens when the panel is on the right.

Please open an issue request with the same example image and your information, so we can figure out what causes it. If do do not want, I'll do it, but it would be best if you do, so you can be informed of any changes and updates on this issue.

As an obvious workaround you may place the panel on the left for the moment to avoid the issue.

PR #578, opened 2020-04-17
Cluster terminals

hanspr:
cluster terminals, take all vertical size available when exploding window terminals.


hanspr:
@gfrenoy 

This is a very small (but useful) change, it could be released with 6.2 as well

Issue #560, opened 2020-04-06
Loki - EL7 - Can't select text from the begining of a line on ssh terminal


popxunga:
Hi ...
This is with latest loki build EL7 with gnome3 (non-classic mode).
When the Connections section is collapsed, in any ssh terminal, I can use the mouse to select some text to copy, starting at the beginning of the line.
When the Connections section is not collapsed, I cannot start selecting from the beginning of the line, because when I put the mouse over the 1st char, the mouse pointer assumes I want to change the width of the Connections section.
As a workaround I have to select the text from the end to the beginning of the line (from right to left). 
Not sure if this happens while using other distro's, but is there anything that can be done to correct it ? I do not remember to have this difficulty before starting to test loki. 
Thanks.

gfrenoy:
Noticed that as well (and found the same workaround ;-p).

Not sure when it has been introduced but fact is that most of the margins has been decreased.  I overlooked that change but I guess it's part of the major theme rework.

At this stage ; I'm wondering if this is a Gtk/theme related issue or if we are doing something wrong with the app...?  My guts say it's a Gtk issue but I don't have an evidence at this stage.

gfrenoy:
This should be fixed in the latest loki build (at least it is for me ;p).

Do you confirm ?

popxunga:
Yep ! ... Confirmed !
Thanks :-)

PR #583, opened 2020-04-22
Fix #581 redesigned statistics to avoid jumps

### LINKED ISSUE #581: Connection panel resizes on terminal selection open
@syco

Documenting case based on #568 

Connection panel will resize when is displayed on the right and you select some connections.

79915792-8648ec00-841f-11ea-8eee-a63135966231

Apparently is produced by the expansion or contraction of the statistics information, when the length of the "Reset" button changes (longer shorter connection names) or the connected time expands, contracts, creates a connection list resize. It also happens when a terminal is open because the panel tries to recover the space.

image

Disabling the show statistics on the Global Preferences, apparently removes the undesired effect. Have not been able to reproduce under other scenarios but the statistics enabled.

Needs further analysis to detect other possible causes.

hanspr:
Fix #581 

The problem happens with left or right panel position. Right panel is more affected by the bug.

The largest problem came from the renaming of the Reset statistics button.

I left the button with a fixed width. And rearranged the messages to be clearer and have the same amount of lines so it would not cause the same jumping effect, vertically if the number of lines changes .

The problem arises when the connection panel is resized to squeeze the tabs panel to its minimum. If you leave no space for the statistics time connection to grow or contract you would get a small jump. Because of the resizing of that panel.

imagen

The new design will stop the large changes seen in the original bug report. But I could not find a way to control the resize, or wrapping to avoided the small changes caused by the time legends. Maybe with your experience in gtk widgets, you can manage to tweak it so it controls that small resize. 

+ I tried to get that label inside a scrolled window, but then a got into a bunch of other troubles. So I dropped that road.
+ Tried to make the legend widget to overflow the text so it would not create a resize, could not do it. (this would be the best solution)
+ Tried to make the widget wrap the text. Could not do it. This would end up making the top info panel resize and make it jump.

I took the opportunity to retouch the Main window to make it cleaner by:

+ Adding a large resize handle that separates the 2 notebooks and have a cleaner view and not an group of thine vertical lines clustering all against each other.
+ Removing border frames decorations
+ Add a border separation to the info window and adding a border to stand out its limits.

imagen

### DIFF
--- a/lib/PACMain.pm
+++ b/lib/PACMain.pm
@@ -468,6 +468,7 @@ sub _initGUI {
     $$self{_GUI}{main}->add($$self{_GUI}{vbox1});
 
     $$self{_GUI}{hpane} = Gtk3::HPaned->new();
+    $$self{_GUI}{hpane}->set_wide_handle(1);
     $$self{_GUI}{vbox1}->add($$self{_GUI}{hpane});
 
     # Create a vbox3: actions, connections and other tools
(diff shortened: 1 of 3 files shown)


hanspr:
Added the line wrap.

This fully stops the connection lists resize.

This would be an extreme example

Peek 23-04-2020 01-07

hanspr:
@gfrenoy 

Check it out, I think this could be added to 6.2

gfrenoy:
As it also fixes #560, I'd definitely want to have it in 6.2 :-)  That one really annoyed me much those last days.

Thanks a lot for fixing two birds in one shot :)

Issue #590, opened 2020-04-28
"Select key bindings to disable from TABs Window" not working for  "alt-p".


misiektw:
**Describe the bug**
Even after "select key bindings to disable from TABs Window" are all set to 'checked' shortcut alt-p still opens Preferences window instead of being passed to terminal.

**To Reproduce**
1. Open preferences, check "ALT" in "Select keybindings to diable". Save and close.
2. Open connection to server, and press alt-p
3. Preferences pop up

**Expected behavior**
alt-p should be passed to terminal. (previous command in midnight commander in my case)

**Environment (please complete the following information):**
 - OS:  Arch
 - Ásbrú Version 6.1.2

hanspr:
Hi @misiektw 

The problem is that you might be disabling in the "Main Preferences" and those settings are useless.

![imagen](https://user-images.githubusercontent.com/1572396/80509513-27173a00-893f-11ea-948b-b34d14ce5cfd.png)

You need to disable in each particular "Terminal settings".

![imagen](https://user-images.githubusercontent.com/1572396/80509563-34ccbf80-893f-11ea-9cc3-c5822f67f9ee.png)

All this logic will be removed with #556 , because is wrong and does not allow to customize keybindings.

I already finished the solution, but I'm still waiting for the main 6.2 version to be released and then I will be able to send the changes to loki for testers.

Follow #556 if you want to know when will the push request is merged and test using the builds from the loki branch. Or do it with the current work around explained above.

misiektw:
I didn't notice this options in individual connections, however it changes nothing. I just tried it. Still alt-p opens preferences instead going into terminal.

However if keybinding remapping will be available soon, then this is pretty much non-issue. I just hope it will be possible to turn them all off....

hanspr:
> I didn't notice this options in individual connections, however it changes nothing.

That keybind is embedded in the button creation and is controlled directly by gtk.

```perl
$$self{_GUI}{configBtn} = Gtk3::Button->new_with_mnemonic('_Preferences');
```

As was not aware of this types of bindings, good to know.

I will send a PR, to remove all this embedded bindings in 6.2, they add even less control to the current keybinds model. And will not be able to be customizable.

Issue #556, opened 2020-04-05
Customize Keybindings


hanspr:
+ Create framework to customize keybindigs for the application
+ Create window to customize actions

I'll take this

Issue #543, opened 2020-03-31
search function : Ctrl-F only works when focus is on connections pane


thijz:
**Is your feature request related to a problem? Please describe.**
Hitting Ctrl-F only opens the search box when the focus is on the connections pane
If you are working in a terminal tab Ctrl-F does nothing

**Describe the solution you'd like**
Would be great to have the search box open no matter where you are working

**Describe alternatives you've considered**

**Additional context**

gfrenoy:
Ctrl-Shift+F is the shortcut you can use when the focus is on a terminal.

thijz:
ah ok,
it seems that this search function is limited to the connection names.  it does not search for ip's

gfrenoy:
This should work better in the loki branch : Ctrl+F is now showing the same search form ; whatever the place you're searching for.  The other request will be handled in #544.

Let us know if this works for you.

Issue #285, opened 2019-11-25
Add hotkeys to commands


park-land:
There are a number of common commands that I currently use the remote commands feature to manage. 

It would be nice if I could assign a hotkey to a command so that I can simply type Ctrl-F1 for example and have the command sent to the terminal, ala MobaXterm macros.

I use the drop down below the terminal which is okay, but switching context from keyboard to mouse slows down what I'm doing and breaks the flow.

gfrenoy:
Would be a nice enhancement, indeed :)

I guess we would want to have a configurable hotkey ?

merikz:
Perhaps a field in the "Remote macros" pane where you can chose hotkey for the entry?

There is this "Disable next keybindings..." in the "Terminal Options" pane that I never really understood what it's good for. But it hints that maybe it's not straight forward to just pick any key combination and assign to a macro.

hanspr:
@gfrenoy I'll take this ; with #556

hanspr:
It has been added to next version 6.3, you can test it now if you install the Snapshot releases (I'm not sure if is already in one now) or the loki releases.

https://docs.asbru-cm.net/General/Installation/

gfrenoy:
You should install the "loki" version to test this further.

PR #600, opened 2020-05-06
Fix556 keybindings

hanspr:
Fix for : #556 #285 #43 #590 #543

__Note: This PR is build on top of restyle-modal-windows__

Added support to customize keybind actions.

Preferences

imagen

Select a row, type your key combination to add to the selected row. `Delete` or `Backspace` disables the corresponding row.

On terminal or Connections list, popup has a show keybindings to list current active keybindings

imagen

imagen

In the Global and Terminal : Remote and Local commands, can have a keybind assigned by placing the cursor on the keybinding field and typing the keybinding.

imagen

Added some additional actions that were not available before. If there are any new actions that should be supported by default, we should add them now.

* Open SFTP connection in SSH terminals `AltCtrl+s`

More actions can be added in the future, but will be less efficient.


gfrenoy:
Brilliant !!!  Thanks a lot, it was a long time waited feature :)

Do I get it right and some of the "legacy" keybindings have changed ?

As an example, I was using:
- Ctrl+F4 to close current terminal, it is now Ctrl+w
- Ctrl+Shift+V to paste selection, it is now Shift+Insert

Is this on purpose ?

hanspr:
Those were duplicated in the original code (so those actions always had 2 keybindigs associated to it), and under this model you can not have 2 different keybinds for the same action, so I had to sacrifice one of them.

I thought those were the more widely used. I have used `Shift + Insert` since I began using my first terminal 30 years ago.

But change it to anything you like, everything can be rebind now, so it makes no difference to me.

gfrenoy:
> I had to sacrifice one of them

Ok fair enough !  I never noticed there were several bindings for the same thing :)  Thanks for heads up.

So this explains ```Shift + Insert``` but ```Ctrl+w``` is not working on 6.2 ; do I get it right ?

> But change it to anything you like, everything can be rebind now, so it makes no difference to me.

Yes it works great and just takes a few seconds, thanks a million !

I was just wondering if we should pay extra attention to keep the exact same keybindings (by default) or if we should "force" a new default keybindings set and assume users will handle it...?

hanspr:
I don't remember how did I came to `Ctrl+w`

> if we should "force" a new default keybindings

That was not the intention, I was never familiar with the original keybindings because I always disabled them. I think I tried to respect what it was on the code (as I read it). But I can not guaranty that I did not make a mistake.

This is the time to fix any mistake; if I made a boo boo

Feel free to set them up how ever you think is best.

What I really would like to re think; is if there are any useful keybindings that do not exists and could be implemented now.

For example, I was not aware that if you right clicked on an ssh terminal there was a menu entry to open an SFTP session using the settings from that same opened terminal. So I added a new keybinding that was not there before. And since then I have used it several times.

I'm sure there could be more little jewels hidden there that I was not aware of and could be added at this time.

hanspr:
Ok, `Ctrl+w` `Ctrl+F4`

The problem with this ones is that I might misunderstand them.

imagen

imagen

`Ctrl+F4` Closes a Tab
`Ctlr+w` Closes a terminal

By code you can read that both do exaclty the same, but are caught at a different level.

So again, I sacrificed the one from the Tabs, thinking that executing it in the terminal should be enough (and it is).

Reimplementing the first one would be the solution. Adding the action to close a "Tab" that it actually means: close the terminal in the "Tab"

hanspr:
Added the commit to recover the duplicated keybinding to `Ctrl+F4`

At least from a users point of view might look differently thinking about closing a "tab" rather than a terminal.

You will have to reset the defaults, to see the change. I did not want to implement a scanning off all the keybindings to detect new ones. Rather when we close the final version, we add them with a different method when needed.

So, right now during testing is: reset to defaults, test, reset to defaults, etc.

gfrenoy:
> ```Ctrl+F4``` Closes a Tab
> ```Ctrl+w```  Closes a terminal
> By code you can read that both do exaclty the same, but are caught at a different level.

Indeed, I see ; thanks for pointing that out !

But ```Ctrl+w``` did not work in the past versions and now they both have the exact same behavior ; I'd like to review this.

I believe they should match the "Disconnect" and "Close Terminal" options we had:

image

Said differently : ```Ctrl+w``` should not close the tab.

Another annoying regression I noticed is that it does not reconnect when you hit "ENTER":

image

Let me do a few more tests as you suggested and see what I can do to smooth the upgrade process.

Thanks again !

hanspr:
>  I noticed is that it does not reconnect when you hit "ENTER"

Missed that one, thanks.

gfrenoy:
Don't know if you noticed but this actually also nicely fixes also #543 :)

hanspr:
> Don't know if you noticed but this actually also nicely fixes also #543 :)

Yes I did know

PR #615, opened 2020-05-13
fix change method, properties not visible

hanspr:
If you edit any connection (new or existing one), and change the method, the newly selected specific method options are not visible until you close the edit window and reopen again.

### DIFF
--- a/lib/PACEdit.pm
+++ b/lib/PACEdit.pm
@@ -301,6 +301,7 @@ sub _setupCallbacks {
 
         &{$$self{_METHODS}{$method}{'updateGUI'}}($$self{_CFG}{'environments'}{$$self{_UUID}});
         $$self{_SPECIFIC}->change($method, $$self{_CFG}{'environments'}{$$self{_UUID}});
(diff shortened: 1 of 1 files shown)


gfrenoy:
I have the vague souvenir that I removed that line in a previous commit ... Is this possible ?

Because it was really on purpose but I cannot exactly remember why :-/

Do you see what I mean or not at all ?

hanspr:
I think that it was removed because it "reshows" or "makes sensitive" any options that you might have hidden or made insensitive during validations of options (mainly in the Network tab where we hide and show elements depending on the connection type).

But now there is an additional line that checks and resets the status of those elements again.

`__checkRBAuth($self);`

So I thinks that it no longer has that __reset__ effect any more.

I think that we have to return the __show all__, and monitor what options in the window get redisplayed again, because of conditions that we implemented. 

And we keep factorizing the "re hiding or re setting the sensitiveness" of any manual manipulations we have done on the UI, to recover its correct status.

Hope I my explanation is not too convoluted.

gfrenoy:
Thanks for the additional information.  I could not find it back so let's give it a try :)

PR #703, opened 2020-10-12
Add win10 instructions to docs

SegiH:
I have added Windows 10 instructions at the top of docs/General/Installation.md

I felt that the Windows 10 section should be at the top of the document because Linux users will be more likely to know that they need to scroll down to find their distro instructions but Windows 10 instructions should be more prominent at the top.

I would also suggest adding this to the Asbru-CM wiki


gfrenoy:
Super cool, thanks a lot Segi !  Do you mind signing the contributor agreement so that I can merge the PR ?  Much appreciated !

SegiH:
I clicked on the sign Contributor License Agreement link and clicked on agree. It showed a message that said "thanks for signing the CLA"  but after it redirects me back to this page, it says that I have not signed the CLA even though it tells me that I have

PR #706, opened 2020-10-16
do not close socket on disconnect only at the time of a full stop

### LINKED ISSUE #502: High CPU utilization when disconnect
**Describe the bug**
When a connection is disconnected, asbru cpu usage is 100%.

**To Reproduce**
1. Open connections
2. Disconnect ssh connection with Ctrl+D
3. Repeat Step 1,2
4. CPU is 100% used by asbru process
5. When asbru main window closed, cpu returns normal

**Expected behavior**
Asbru process should not use cpu 100%

**Screenshots**
Screenshot CPU usage
[https://i.postimg.cc/qMRHC9LS/Screenshot-from-2020-03-24-11-55-19.png](url)

 - OS: Ubuntu 16.04
 - Version 6.1.0rc2

**Additional context**
I am not sure how many times connect/disconnect

hanspr:
Regarding #502 

I could stop the sysread errors and the high CPU, by commenting the lines that close the socket when a disconnection is issued.

The high cpu and sysread errors on closed file handle, are due to the sockets being closed when other parts still try to read from it.

I have not seen other side effects on the rest of the application by not closing the socket during a disconnect. I could not see oprhaned sockets or some other error introduced by this.

In ubuntu 18.3
As much as I try to follow the code I could not come up with a clear understanding of when the socket creates a high cpu. There are cases where it never happens others where it does randomly.

So you could give it a test and if you think this does not creates other unseen problems by me, it should fix #502 .

### DIFF
--- a/lib/PACTerminal.pm
+++ b/lib/PACTerminal.pm
@@ -1282,9 +1282,9 @@ sub _setupCallbacks {
             $self->_showEmbedMessages();
         }
 
-        if (defined $$self{_SOCKET_CLIENT}) {
-            $$self{_SOCKET_CLIENT}->close();
-        }
+        #if (defined $$self{_SOCKET_CLIENT}) {
+        #    $$self{_SOCKET_CLIENT}->close();
+        #}
         if (defined $$self{_SOCKET_CLIENT_EXEC}) {
             $$self{_SOCKET_CLIENT_EXEC}->close();
         }


gfrenoy:
I see this socket is closed at different places in the code ... Won't we have the same issue elsewhere ?  Why removing that one and not the other ones ?

hanspr:
The only place that is the correct one is in the sub stop {} method.

I do not know the other places you mention, it would have to be tested in those cases, looking a the events that trigger that piece of code. Currently I have't seen issues.

gfrenoy:
> I do not know the other places you mention,

In "destroy" and in "child_exited" ; look for `{_SOCKET_CLIENT}->close`.

gfrenoy:
In "child_exited", shouldn't we add:

`
$$self{_SOCKET_CLIENT} = undef;
`

after we close the socket ?  So it will not try to close it in "stop" ?

hanspr:
Destroy sound reasonable, is when we completly destroy the PACTerminal object.

Child Exited is where I commented.

`$$self{_SOCKET_CLIENT} = undef;`

Do not think is necessary, is part of the object and by then would be completely wiped out.

gfrenoy:
> Child Exited is where I commented.

Ach.  This is where I was lost ; thought it was in "close" !  Sorry.

PR #725, opened 2020-11-07
fix vte_feed_child detection

syco:
Signed-off-by: Syco <alberto.rinaudo+dino@gmail.com>


syco:
ignore this, looks like it's more complicated.

PR #726, opened 2020-11-07
update vte_feed_child detection

syco:
Signed-off-by: Syco <alberto.rinaudo+dino@gmail.com>

### DIFF
--- a/lib/PACMain.pm
+++ b/lib/PACMain.pm
@@ -4815,6 +4815,7 @@ sub _setVteCapabilities {
     eval {
         local $SIG{__WARN__} = sub { die @_ };
         $vte->feed_child('abc', 3);
+        1;
     } or do {
         $$self{_Vte}{vte_feed_child} = 1;
     };
--- a/lib/PACMain.pm
+++ b/lib/PACMain.pm
@@ -4813,8 +4813,8 @@ sub _setVteCapabilities {
     #  See https://bugs.launchpad.net/ubuntu/+source/ubuntu-release-upgrader/+bug/1780501)
     $$self{_Vte}{vte_feed_child} = 0;
     eval {
+        local $SIG{__WARN__} = sub { die @_ };
         $vte->feed_child('abc', 3);
-        1;
     } or do {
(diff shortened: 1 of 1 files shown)


gfrenoy:
Thanks!

For the other distros, don't we need to keep the `1;` so the `eval` succeeds ?

syco:
The 1 doesn't make any difference in my tests, here's how I'm doing it:
```perl
use strict;
use warnings;
use Gtk3 -init;
use Glib::Object::Introspection;

Glib::Object::Introspection->setup(basename => 'Vte', version => '2.91', package => 'Vte');

my $terminal = Vte::Terminal->new();

my $str = "# do nothing";
use bytes;
my $b = length($str);
my @arr = unpack ('C*', $str);

eval {
  local $SIG{__WARN__} = sub { die @_ };
  print("Test 2 args: ");
  $terminal->feed_child($str, $b);
  print("OK 2\n");
} or do {
  print("ERROR, should be 1\n");
};

eval {
  local $SIG{__WARN__} = sub { die @_ };
  print("Test 1 args: ");
  $terminal->feed_child(\@arr);
  print("OK 1\n");
} or do {
  print("ERROR, should be 2\n");
};

exit(0);
```
```bash
$ perl terminal.pl 2>/dev/null
Test 2 args: ERROR, should be 1
Test 1 args: OK 1
```

gfrenoy:
I would guess the `print` operation is returning something equivalent to 1 so that it works.  Can you try without anything after `feed_child` ?

As a matter of fact, this is what we are doing in `_setVteCapabilities`:

 ```perl
    $$self{_Vte}{vte_feed_child} = 0;
    eval {
        $vte->feed_child('abc', 3);
        1;
    } or do {
        $$self{_Vte}{vte_feed_child} = 1;
    };
```

On my system, `vte_feed_child` should be `0` but if I remove the line `1;`, it selects `1` for `vte_feed_child` and I'm having all kinds of little glitches...

syco:
Yes, if I remove the print statement it changes behaviour.
I've added a second commit for it.

PR #743, opened 2020-12-09
Fix bug clipboard utf8

hanspr:
If text is selected on the terminal, and that text has utf8 characters, the text will loose characters at the end of the string because the length property is passed as a size of the string and not size of bytes

For example the following text selected in the terminal

`enable programmable completion features (you don't need to áéíóúñÑ enable`

Would be copied as this in the clipboard

`enable programmable completion features (you don't need to áéíóúñÑ`


gfrenoy:
Thanks a lot !  Great finding :)

PR #762, opened 2021-02-08
Add random sock5 port to connctions and add custom command shortcur

syco:
I have 2 changes packed in here:

1st small change: remove old netstat and use perl library. In OpenSUSE I don't have netstat anymore, but ss. To avoid having to detect what binary is available I preferred to implement a native perl library.

2nd: add a new option to add a random sock5 port to all connections and allow a custom command to use it:
I often have the need to open a browser tunnelled to a remote connection, with hundreds of connections configured is unfeasible to remember what `-D` ports I've already used, hence the need to add a dynamic random port.
The command I use for testing is `TMP_HOME="$(mktemp -d)"; HOME="${TMP_HOME}" /usr/bin/chromium --incognito --proxy-server="socks5://localhost:<<PORT>>" --proxy-bypass-list="<-loopback>"; rm -Rf "${TMP_HOME}"`

This opens unlimited temporary browsers with no plugins or other customizations, in incognito mode, tunnelled to the remote connection.
Note that if you don't use the temporary home, or have chromium already opened, the proxy option doesn't take effect.
I, for instance, use firefox as my main browser and chromium for this temporary needs.

image

image

Please let me know if you have any suggestion, I've been using it for months, but I'm sure it could be further improved.
Thanks

### DIFF
--- a/doc/Manual/Connections/SSH.md
+++ b/doc/Manual/Connections/SSH.md
@@ -40,6 +40,7 @@
 
 ![](images/ssh2.png)
 
++ __Open sock5 tunnel__ : Append '-D <<random port>>' to this ssh connection and use the port to open an application using <<PORT>> as a parameter.
 + __Programmatically send a string__ : Send a specified regex expression every selected seconds to the terminal.
 + __Prepend command__ : Add this command before the ssh command connection string.
 + __Start next script when connection is launched__ : * Pending
--- a/lib/PACEdit.pm
+++ b/lib/PACEdit.pm
@@ -332,6 +332,12 @@ sub _setupCallbacks {
         }
     });
 
+    # Capture 'sock5 tunnel' checkbox toggled state
+    _($self, 'sock5TunnelActive')->signal_connect('toggled' => sub {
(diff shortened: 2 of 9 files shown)


gfrenoy:
Waw, this is really great !  Thanks a lot for your contribution.

The remarks I have for tonight:

1) I actually initially wanted to use native perl to do this but could not find how to achieve this so I used netstat.  Removing that dependency will be great.

2) that "automatic" sock5 proxy is really great.  I'm just wondering what would be the default ?  I believe we don't want to have chromium as a dependency ?

3) some documentation about that feature will be required / useful

4) i'd like to schedule this for 6.4

Thanks again !

syco:
I was reading the doc to see where to add this and I think the "Local Macros" mostly covered this functionality already.
I'll see how it can be done better and push a new pull request shortly.

gfrenoy:
So we should close this one ?  It is fully replaced by #771.  Do I get it correctly ?

syco:
Yes, it can be closed..

PR #771, opened 2021-02-15
Add random socks5 port to connctions and add related variable to use in local commands

syco:
Following the previous epiphany on pull request #762 I've reimplemented this function as a local command.
The functionality I had intended is still the same, and I'm happy with both implementations, but I think this way it integrates better with what's already there.

Just as an example, the command I use with this new variable is: `bash -c 'TMP_HOME="$(mktemp -d)"; HOME="${TMP_HOME}" /usr/bin/chromium --incognito --proxy-server="socks5://localhost:<SOCKS5_PORT>" --proxy-bypass-list="<-loopback>"; rm -Rf "${TMP_HOME}"'`

Thanks

### DIFF
--- a/doc/Manual/Connections/SSH.md
+++ b/doc/Manual/Connections/SSH.md
@@ -40,6 +40,7 @@
 
 ![](images/ssh2.png)
 
++ __Open socks5 tunnel__ : Append '-D <<random port>>' to this ssh connection and use the port to open an application using <SOCKS5_PORT> as a parameter.
 + __Programmatically send a string__ : Send a specified regex expression every selected seconds to the terminal.
 + __Prepend command__ : Add this command before the ssh command connection string.
 + __Start next script when connection is launched__ : * Pending
--- a/doc/Manual/Substitutions.md
+++ b/doc/Manual/Substitutions.md
@@ -20,6 +20,7 @@ The general format of the mask is `<[(Name|Type):]VALUE>`.
 |USER              |Connection USER                    |
 |PASS              |Connection Password                |
 |UUID              |Connection UUID                    |
+|SOCKS5_PORT        |Socks5 tunnel port                  |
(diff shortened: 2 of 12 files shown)


gfrenoy:
Alberto, as this option would only work with SSH terminals, shouldn't we move the option from the "Advanced" panel to the specific SSH options panel ?

Or said differently : does that option has any sense for VNC or RDP terminals ?

syco:
Yes, it probably should be in the SSH advanced options.
But that panel is built programmatically and current values are in the 'options' variable that needs to be exploded and parsed.
And I didn't find a variable I could reference to copy how it works.

gfrenoy:
Indeed, nothing is really foreseen for Ásbrú specific options.

I tried the following (this could be extended later to other needs):
 - using a dummy option "#Ásbrú#RST#  (RST = Random Socks Tunnel)
 - any dummy option is cleaned up when starting `asbru_conn` (so it is not considered)
 - replaced `socks5 tunnel active` by a check for that dummy option in the connection options

You might need to reconfigure your connection ; there is no migration path foreseen.

Let me know what your thoughts...

syco:
You made me do work... joke aside...
that works, but we could also add another configuration parameter along side with 'options' (I called it 'connection options') that is an array, and contains specific configurations for every connection method.
This way we can access it directly without the parsing functions.
In time all options could be migrated to this new array.
The only wrinkle is that it needs to be implemented, even as an empty array, for all connection methods.

Unless I forgot something, this is it: [https://github.com/asbru-cm/asbru-cm/compare/loki...syco:loki-feature-socks-tunnel](https://github.com/asbru-cm/asbru-cm/compare/loki...syco:loki-feature-socks-tunnel)

gfrenoy:
Yeah, indeed, way much better :clap: :clap: :partying_face:  !!!

Thanks a lot :)

Let me merge this into loki.

PR #772, opened 2021-02-15
Change max baud rate for using above 1 Mbps on serial(using cu)

nlimbo76:
Check please.
Thank you!


KlaasT:
Thank you very much for your contribution. It would be great if you could sign our CLA please.

In my opinion this change shouldn't do any harm and only changes the max value. Would be fine for me.

What do you think? @gfrenoy

PR #784, opened 2021-02-24
Do not mention yaourt

micwoj92:
yaourt has been abandoned for several years now. It lacks many features and has many security issues.

I changed the mention so now it mentions [the manual install process](https://wiki.archlinux.org/index.php/Arch_User_Repository#Installing_and_upgrading_packages). I do not believe that any AUR helper should be promoted. It is up to user which one they want to use.

Also changed line so it says that the instructions are for pacman-based distros. This is because pacman contains the makepkg tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,10 +56,11 @@ To do so, execute the following commands:
   $ sudo dnf install asbru-cm
   ````
 
-- Arch / Manjaro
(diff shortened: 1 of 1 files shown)


gfrenoy:
Thanks a lot for your expert advise and great suggestion !  Very much appreciated :)

PR #802, opened 2021-04-08
remove all embedded which checks and replace them with system

syco:
Hi,

I have a problem with the detection of xfreerdp, for some reason https://github.com/asbru-cm/asbru-cm/blob/master/lib/PACUtils.pm#L443-L446 doesn't work for me.
But because I find the way `which` is used very ugly and inconsistent, I've decided to change all of them with `system("") eq 0`.
This way the variables `my $PERL`, `my $xfreerdp`, `my $xtightvncviewer`, etc.. now contain a value already sanitized that can be used as is.

### DIFF
--- a/lib/PACScripts.pm
+++ b/lib/PACScripts.pm
@@ -41,7 +41,7 @@ use Storable qw (dclone nstore_fd);
 eval {require Gtk3::SourceView2;};
 my $SOURCEVIEW = ! $@;
 
-my $PERL = `which perl 2>&1`;
+my $PERL = (system("which perl &>/dev/null") eq 0);
 
 # GTK
 use Gtk3 '-init';
--- a/lib/PACUtils.pm
+++ b/lib/PACUtils.pm
@@ -368,10 +368,9 @@ sub _getMethods {
         $THEME_DIR = $theme_dir;
     }
 
-    `which rdesktop 1>/dev/null 2>&1`;
-    my $rdesktop = $?;
+    my $rdesktop = (system("which rdesktop &>/dev/null") eq 0);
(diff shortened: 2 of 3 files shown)


gfrenoy:
This is surprising, backticks and system are basically very similar (see [here](https://stackoverflow.com/a/800034)).  I'm surprised to read only one of the call is not working fine ?!

That being said, having a clear and consistent way for this is indeed great !  Thanks for that.  Maybe we should have a dedicated helper function in PACUtils ?

syco:
I didn't try ALL of them (for example I don't have mosh or autossh) but it works for perl/vncviewer/telnet/ncat and doesn't work for xfreerdp.
It might be related to permissions, but I don't see any major difference here:
```bash
$ which xfreerdp
/usr/bin/xfreerdp
$ ls -la /usr/bin/xfreerdp
-rwxr-xr-x 1 root root 686824 Apr 20 06:33 /usr/bin/xfreerdp
$ which vncviewer
/usr/bin/vncviewer
$ ls -la /usr/bin/vncviewer
lrwxrwxrwx 1 root root 27 Dec 21 11:39 /usr/bin/vncviewer -> /etc/alternatives/vncviewer

```

It could be a function, but it's also one line of code, is it worth moving it in its own function?

gfrenoy:
> backticks and system are basically very similar (see [here](https://stackoverflow.com/a/800034))

Ok I read the documentation again and there is a major difference between both approaches : `system` returns the exit status of the command while backticks return `STDOUT` of the command.  Both put process exit code in `$?`.

So all in all, it *should* work but your "cleanup" is indeed a nice step forward to have a more consistent way of managing this.

Thanks a lot :)

Issue #742, opened 2020-12-04
Show available presets for ssh port forwarding when connected


BigDollarini:
**Is your feature request related to a problem? Please describe.**
When connected to a host it's difficult to remember local and remote port mapping via ssh.

**Describe the solution you'd like**
While connected to a host a sidebar/side window/menu option/listing should show preset forwarding on local and remote ssh tunnel ports available.

**Describe alternatives you've considered**
A good example would be a older product called XShell which also showed realtime stats on sockets and up/down total bytes. All I need to see is what is available to connect to locally.

**Additional context**
Converting from PAC-VS installed via snap was a challenge, maybe some new doc's on that too !!! but that's another request.

gfrenoy:
Very cool idea, thanks for the suggestion !

kjuanman:
Hi, I use asbru every day and I think I could help with a patch for this feature.

May be instead of a sidebar, the info could be showed in the status bar?

kjuanman:
I pushed a first version of the proposed feature. It add the Forwarding Ports information in status bar in ssh connections:
![image](https://user-images.githubusercontent.com/5916928/122312724-36fb3000-ceeb-11eb-8987-b9e70858050e.png)

I could add a option in preferences to opt-out/opt-in in Look & Feel.

gfrenoy:
@kjuanman thanks for your contribution, it works fine :-)

Do you think we can make it a bit "user friendly" / "human readble" ?

I mean, if you don't know the `ssh` options (`-L`, `-D`, etc.), it is not easy to read/understand.

kjuanman:
> @kjuanman thanks for your contribution, it works fine :-)
> 
> Do you think we can make it a bit "user friendly" / "human readble" ?
> 
> I mean, if you don't know the `ssh` options (`-L`, `-D`, etc.), it is not easy to read/understand.

I send a Pull Request based in that suggestion: 
It add options in preferences to:
-Show port forwarding information in status bar
If yes:
-Use full names (Local Forwarding, Remote Forwarding, Dynamic Forwarding)
-Use abreviatures (-L, -R, -D)

Regards

gfrenoy:
@kjuanman Great contribution ! Thanks a lot.

When there is no port forwarding, I see the following error messages on the command line:

```
Use of uninitialized value $forward_ports in substitution (s///) at /path/to/asbru/asbru-cm-loki/lib/PACTerminal.pm line 2455.
[...]

PR #830, opened 2021-07-09
Fix use of uninitialized value $forward_ports (#742)

kjuanman:
Fix warning "use of uninitialized value $forward_ports" when there is no forwarding ports https://github.com/asbru-cm/asbru-cm/issues/742#issuecomment-875031745


gfrenoy:
Thanks :)

PR #848, opened 2021-09-29
[Feat] Partial Configuration Import

henry701:
Enables the capability to import partial configuration "diffs" from `asbru.yml`. Currently no export capability is provided, but this feature is useful to distribute "flavors" of Ásbrú with only some configurations differing from the defaults.

Notably, though, this adds a dependency on `Hash::Merge::Simple`.


henry701:
An example of a file that can be imported in this way is as follows, for only changing the fonts:
```yaml
---
__PAC__EXPORTED__PARTIAL_CONF:
  defaults:
    info font: Ubuntu Mono 12
    terminal font: Ubuntu Mono 12
    tree font: Ubuntu Mono 12
```

henry701:
Accidental close because of accidental branch delete. Oops.

henry701:
The recent force-push added some code to import configurations on first start (or when configuration is otherwise absent) from file `$RealBin/vendor/asbru-conf-default-overrides.yml`, if it exists.

PR #857, opened 2021-10-27
Add the PostPend command in order to use ChromaTerm

teicors:
Hi, this i s the part of the code to start to implement a "Postpend" command ideally created for ChromaTerm, in order to colorize the output of a SSH session (like MobaXterm).

In the postpend command you have to put " | ct "

There's a now problem with SFTP session: in my mind the string used to connect by SFTP is the same as the connection string used for the SSH.


gfrenoy:
Thanks a lot !  Code has been merged into `loki`, let's test this now :)

PR #860, opened 2021-11-01
[Feat] Vendor default configuration overrides

henry701:
Added some code to import configurations on the first start (or when the configuration is otherwise absent) from file `$RealBin/vendor/asbru-conf-default-overrides.yml`, if it exists.

More reasoning behind the feature can be seen in [the new documentation section](https://github.com/henry701/asbru-cm/blob/loki-feat-vendor-defaults/doc/General/Redistribution.md).

Unlike [the original PR](https://github.com/asbru-cm/asbru-cm/pull/848), this does not include any additional dependencies, because there is no need to merge with existing configuration items when there are no existing configuration items in the first place (starting Ásbrú without preexisting configuration).

### DIFF
--- a/doc/General/Redistribution.md
+++ b/doc/General/Redistribution.md
@@ -1,0 +1,24 @@
+# Redistributing Ásbrú Connection Manager
+
+When redistributing Ásbrú Connection Manager (e.g acting as a vendor), you can customize some aspects of it to better accomodate the specific distribution/environment you are distributing it on.
+
+## Customizing Default Configuration
+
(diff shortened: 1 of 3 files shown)


gfrenoy:
Thanks a lot for your help Henrique !

Don't we need to add an entry in `mkdocs.yml` to see you new documentation page ?

henry701:
Added it to the suggested section. I didn't know the index was manual, I thought it was folder-based, my bad.

PR #863, opened 2021-11-02
Tunnels once

hanspr:
If you open a terminal that forwards ports or creates tunnels, the first time you open that terminal you get your clean simple loggin.

imagen

But if you open a second terminal, you get warnings that the ports are in use.

imagen

The simplest way to avoid this is to delete the tunnel options from the command when the tunnels are already open. Not a very fancy solution but works.


gfrenoy:
Thanks !  I do see that message often and it is indeed not a very nice one :-/

I was just wondering if it's OK to completely ignore the error...?  I mean: the message is not nice and could be more clear but it's a "real" error ... The tunnel could not be established for this connection so it means that if you close the other connection, the tunnel will be closed as well.

Should we at least show a message that is displayed if you enable the "--verbose" mode ?

hanspr:
Well I think I could add a yellow message on the terminal , saying "Skipping tunnel openings, already open in a different terminal", at least it would be a nicer message and a reminder.

If you like this I could implemented and send another commit

hanspr:
Would look similar to this

imagen

gfrenoy:
> Would look similar to this

Looks great !  Thanks :+1: 

Maybe showing the port to give a clue about which one we're talking about ?  And "opening" has only one "n", I would think.

hanspr:
new message

imagen

gfrenoy:
> new message

Wonderful, I do see this message nearly everday so I'll think about you very often :)

Thanks a bunch !

PR #865, opened 2021-11-02
Fix 729

hanspr:
Save last copied array ref in new reference and use for next paste action.

Only implemented for copy


hanspr:
The decoding is over the paths not over the content of the files, if now fails to find the correct location of the database file, will have the same problem with the key path. The problem is that at the moment of writing the content to the file the Yalm library encodes everything to utf8, and those values are already in utf8, so they get double encoded.

PR #869, opened 2021-11-06
Paste

hanspr:
Added right click paste multiple times, I'm a keyboard driven user, barely use the mouse.


hanspr:
This image proves how clumsy I'm with the mouse

Peek 06-11-2021 14-25

Issue #870, opened 2021-11-08
Tray Icon in Unity


toptnc:
**Is your feature request related to a problem? Please describe.**
Since GTK3::Appindicator does not exists, tray icon in Unity has been disabled

**Describe the solution you'd like**
Other GTK3 Perl apps like Shutter have solved the problem without needing a Gtk3::Appindicator module.
https://github.com/shutter-project/shutter/blob/master/bin/shutter#L647

It seems like Gtk3::StatusIcon may work in Gnome Shell and Unity

toptnc:
I think I have done it. 

Patch is attached.


[indicator-unity.diff.gz](https://github.com/asbru-cm/asbru-cm/files/7513274/indicator-unity.diff.gz)

gfrenoy:
Thanks a lot for your help on this !  That'd be great if we can manage this properly !

Let's keep this open until we finalize the PR...

gfrenoy:
Fernando, it seems that this is causing some side issues and I'd like to enable the use of `AppIndicator3` only when required and use the default tray icon as the first choice.

Can you give me a bit more information about your environment:
- which distro (Ubuntu I guess)
- which version (Unity is not used anymore by default...)
- value of environment variables `XDG_SESSION_DESKTOP`
- are using using GNOME extensions like Unite or Top Icons ?
- what is shown when you run `./asbru-cm` from the commandl line (especially the detected environment)

Thanks.

toptnc:
Distro: Ubuntu 20.04
Unity Version: 7.5.1+20.04.20211026
XDG_SESSION_DESKTOP=unity
Not, i'm not using Gnome extensions since I don't use Gnome

This is my output when I run asbru-cm from the command line

***@***.***:~$ /opt/asbru/asbru-cm
INFO: Ásbrú Connection Manager 6.4.0 (asbru-cm) started with PID 298610
INFO: Language detected : es_ES.UTF-8, forcing to en_US.UTF-8
INFO: Desktop environment detected : other
INFO: Config directory is '/home/fsancho/.config/asbru'
GLib-LOG **: unsetenv() is not thread-safe and should not be used after
threads are created at
/usr/lib/x86_64-linux-gnu/perl5/5.30/Glib/Object/Introspection.pm line 67.
INFO: Used config file '/home/fsancho/.config/asbru/asbru.nfreeze'
INFO: Virtual terminal emulator (VTE) version is 0.60
[...]

PR #871, opened 2021-11-12
Enabling Tray Icon in Unity (#870)

toptnc:
Little and dirty patch to re-enable Unity Tray icon.

### DIFF
--- a/lib/PACMain.pm
+++ b/lib/PACMain.pm
@@ -2172,7 +2172,7 @@ sub _setupCallbacks {
             } elsif ($$self{_CFG}{defaults}{'when no more tabs'} == 2) {
                 #hide
                 if ($UNITY) {
-                    $$self{_TRAY}{_TRAY}->set_active();
+                    $$self{_TRAY}{_TRAY}->set_status('active');
                 } else {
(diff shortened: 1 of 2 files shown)


gfrenoy:
Thanks a lot for your proposal Fernando!  Great job :)

Before merging this, I have a few remarks/requests:

1)
I've removed the "dead" code ; no reason to keep it commented out if it's not used.

2)
I was wondering if `set_status` is backward compatible ?  I mean `set_active` and `set_passive` was working fine ... What is the reason for the change ?

3)
In lib/PACTrayUnity.pm, do we really need to "die" if `AppIndicator` is missing ?  Or this just a warning ?

4)
Would you mind re-publishing this to our development branch (`loki`).  As this is a change that will require more tests, we should not merge it directly to the `master` branch but first on `loki`

toptnc:
1: No problem
2: set_status gave me errors (function not found) so I used set_status instead.
3: Your are right, maybe a warning is just enough, but I don't know the implications. I haven't tested it
4: No problem. I'll make a new pull request against 'loki'

I'm not an expert at all. I was looking into "Shutter" code trying to follow the same method they followed. Feel free to change all you need.

gfrenoy:
Continued with #872

PR #872, opened 2021-11-13
Re-enable Unity tray icon in GTK3

toptnc:
New pull request against 'loki' branch.

As I said before, I'm not an expert in Perl or in GTK so change everything you need. This patch re-enables Tray Icon in my Unity Desktop, and works for me. That's all.

Regards.


gfrenoy:
Thanks really great Fernando :)  Your PR has been merged.

For the records:
- Don't use tabs (but spaces) to indent (see our [coding standards](https://github.com/asbru-cm/asbru-cm/blob/master/CODING_STANDARD.md))
- If the module does not "die", we need to return an error otherwise PACMain will not see the module could not be loaded

PR #873, opened 2021-11-13
fix #755

### LINKED ISSUE #755: "Linux" type function keys support as putty does
I connect to a Linux server that has a somehow strange configuration to run some old ncurses programs which use function keys.
I use Linux version of putty with "Linux" function keys setting to be able to use function keys.
Can we have the same ability in asbru?

See the putty mentioned setting here:  https://imgur.com/a/VZxAE5C

hanspr:
This is the implementation for terminal emulation F keys #755

To test, on terminal:  Ctrl+V[key] shows the key sequence.

So, you type Ctrl+V and then afterwards the key you want to see its sequence

Function keys assigned to asbru actions  in keybindings will be processed without being sent to the terminal, as expected (F4, F11).

If a user wants to pass the key to the terminal, it has to unbind that action in keybindings.

**default xterm : testing: F1,F2,F3,F4,F5**

Peek 13-11-2021 11-35

**Change of emulation for that terminal and then reopen, testing: F1,F2,F3,F4,F5**

Peek 13-11-2021 11-36


hanspr:
will try to remove noise from PACMain.pm, that has nothing to do with this

PR #875, opened 2021-11-13
Fix #662

### LINKED ISSUE #662: Loki EL7 - rdesktop session loose focus once Ctrl+F is pressed
**Describe the bug**
While in a rdesktop session, when Ctrl+F is pressed the focus moves here:
image
To bring the focus back to the rdesktop session, I press the button Set Keyboard focus at the bottom.
After that the keyboard acts like crazy and is simply unusable. 
I do have to close the rdesktop session and open it again to get the keyboard working.

**To Reproduce**
Open any rdesktop session.
While in the rdesktop session open a text file for edition.
Press Ctrl+F.

**Expected behavior**
The focus should remain in the rdesktop session and activate the Text Search find dialog.

**Screenshots**
I see this messages upon opening the rdesktop session (maybe a different issue):
image
These are the settings in my rdesktop session:
image
Tried with Keep WM key bindings as well and the behavior was the same. 

**Environment (please complete the following information):**
 - OS: EL7.8
 - Ásbrú Version [Loki]

Thanks for your support.

hanspr:
fix #662 

Ignore pacmain find if there are opened terminals of any kind. When none is opened honour find requests on the connection tree.

If any kind of terminal is opened. Ignore requests to find, unless the connection tree has focus.


hanspr:
Forget it, I have to re think this one a little more.

PR #881, opened 2021-11-15
Delay user expect

hanspr:
If the prompt is detected a little bit early, even by a character. The user expect sequences will be sent early and they would appear before the actual prompt.

This delay, removes this issue.

It usually happens when the login prompts have colors, because the prompt regex confuses terminal character sequences as part of the prompt.


gfrenoy:
Shouldn't we have a parameter for this ?

I mean: why should I way 1/4 of second every time I open a terminal if this is useless for me ...?

hanspr:
I think this is something that is happening to everyone, but because does not causes any real conflict no one pays attention to it.

I never did a test of how low could you go just to avoid the issue. Apparently with 0.05 1/20th of a second is enough.

If still you do not think is necessary or 1/20th still to much delay, dismiss it.

PR #882, opened 2021-11-16
Fix expect

hanspr:
Hi @gfrenoy , an apology for #881, it was a useless line. Only made an apparent fix but it was cosmetic and it did not work on slow connections.

This is the real correction. It took me some time to debug it. I went back all the way to 2015 and that section has 2 bugs from the very beginning.

https://github.com/asbru-cm/asbru-cm/blob/fc25e6648bda43e61efe9c584e08f2e27c0326b6/lib/pac_conn

The flag "avoid first expectation" should start with false, and there the a jump flag set to true , was missing. The generic mode had the initial flag correct, but both had the jump flag missing when the user marks a successful regex to jump to.

I guess must of us do one expect or sequential ones without the on success on, fail, so the bug never came out, and we all were used to see the command being sent beforehand but ignored it.

So this should correctly address the second issue reported at #876


gfrenoy:
Great finding :)  So does that mean we need to revert the [last commit](44fa3fe3bccc0dd1b23ca656b2649a878acf7daf) ?

hanspr:
> So does that mean we need to revert the last commit ?

Or create a new commit deleting it, I guess. You are the git expert :wink:

gfrenoy:
Don't understand why the GitHub diff does not show that you reverted the other commit but let's go for it ; I'll make another commit after this one if necessary...

PR #885, opened 2021-11-20
Fix #816

### LINKED ISSUE #816: Various issues after upgrade of Crypt-CBC
**Describe the bug**
import connections fails

**To Reproduce**
1. Go to My Connections
2.  left click, select the option import connections
3.  stays permanently on, please wait, it is being imported.

**Expected behavior**
import connections, only a few are imported randomly 

**Screenshots**

**Environment (please complete the following information):**
 - OS:  archlinux
 - Ásbrú from master git version  (6.3.2.r6.g036cf2e-1)
**Additional context**

```
WARN: Adding new keybinding [Shift+Insert] for [Terminal]/[Paste selection into terminal]...
Use of uninitialized value $parent_uuid in hash element at /opt/asbru-cm/lib/PACMain.pm line 2761.
Use of uninitialized value $parent_uuid in hash element at /opt/asbru-cm/lib/PACMain.pm line 2761.
WARNING: The key derivation method "opensslv1" is deprecated. Using -pbkdf=>'pbkdf2' would be better.
Pass -nodeprecate=>1 to inhibit this message.
 at /opt/asbru-cm/lib/PACUtils.pm line 2718.
*** unhandled exception in callback:
***   Salt must be exactly 8 bytes long at /opt/asbru-cm/lib/PACUtils.pm line 2718.
***  ignoring at /usr/share/perl5/vendor_perl/Gtk3.pm line 572.
```
well it seems to be a general problem, maybe with some update of the libraries, if I want to save some changes in the preferences they are not saved.

hanspr:
I stumble with this issue while testing the prompt regex in fedora, a new installation will not save config file and the salt must be 8 bits long shows up.

I managed to remove the error by validating that the strings to encrypt are not empty, apparently empty strings create the error.

Change should not create conflicts with current installations even if empty strings were encrypted before, they should return an empty string.

The other point where it could fail is in PACConfig.pm when they set an empty gui password, I did not move anything there, could be a good test.


hanspr:
How do I fix the  #885 issue that shows with all my Push requests, what happened there ?

hanspr:
Nop, it does not work, sorry

Issue #858, opened 2021-10-31
"Save and Close" button does not save configuration to a file and does not close configuration window


alexus-m:
**Describe the bug**
"Save and Close" button successfully applies changes but does not saves configuration to a file and does not close configuration window neither connection's configuration nor general preferences.

**To Reproduce**
It is the first installation of Asbru-CM. There was neither configuration folder "\~/.config/asbru/" nor legacy "\~/.config/pac*".
1. Start Asbru-CM.
  Open "Edit connection" or general "Preferences" window.
2. Make some configuration changes then press "Save and Close" button.
3. The configuration window left opened on screen:
  The configuration changes are successfully applied to a current process but configuration does not saved to a configuration file and configuration window is not closed. The window can still get closed via neighbour "Close without saving" button.)
There are some debug records appear in terminal window:
DEBUG:KEEPASS: Update usage
*** unhandled exception in callback:
***   Salt must be exactly 8 bytes long at /usr/share/asbru-cm/lib/PACUtils.pm line 2713.
***  ignoring at /usr/share/perl5/vendor_perl/Gtk3.pm line 572.

4. If try to close Asbru-CM window via Ctrl-Q or via "Quit" button then asbru-cm process left running in background and "new" start of Asbru-CM shows up all the configuration I made and "saved".
  If I close the main window via window management "X" button then asbru-cm process disappears and new start begin with clean configuration.
  There is no asbru.yml file appears in ~/.config/asbru/. But some files are there, e.g. 
asbru.pcc
asbru_stats.nfreeze
asbru.yml.gui
asbru.yml.tree
**autostart
bak
screenshots
scripts
session_logs
sockets
tmp**

**Expected behavior**
"Save and Close" button should save configuration to a file and then close the configuration window after changes have been saved.

**Screenshots**
No need.

**Environment (please complete the following information):**
 - OS: Fedora 35
 - Ásbrú Version: loki (asbru-cm-6.4.0~1635596963-1.fc34.noarch)

gfrenoy:
Likely related to #816, strange that it does not work for you... Can you confirm which version of Crypt::CBC ?

alexus-m:
> Likely related to #816, strange that it does not work for you... Can you confirm which version of Crypt::CBC you are using (likely 3.04) ?

Hi @gfrenoy ! Thanks a lot for your response!
$ rpm -qa | grep -i perl.*cbc
perl-Crypt-CBC-3.04-4.fc35.noarch

[...]

PR #886, opened 2021-11-21
Fix missing salt

hanspr:
This is a very brut workaround the issue if the missing salt.

I honestly do not know why it gets lost, probably at some point the new CBC version removes it, and haven't found where, but by resetting it before using the encryption and decryption solves de problem.

I also changed the logic on how to remove the lock password because it was not working, so I simply used the same way is done in the other sections where the lock password is tested.

I managed to make a test of this on a fedora virtual machine where a fresh asbru would no save the configuration with, showing the 8 bytes long salt error, reported on some issues.

I may work in case like : #858


gfrenoy:
Oh yeah, indeed, good workaround !  I feel some dumb that I did not think about this :-/  Thanks a lot !

PR #902, opened 2021-12-19
Feature: Only Programatically Send String when Idle

henry701:
Add "Only when Idle" modifier to the "Programmatically Send String" function of Advanced Parameters connection settings.

The main use-case for this is when your send-string is meant to bypass firewalls that ignore TCP Keep-Alive and kill idle sessions anyway, however could interfere with normal terminal usage when actively using the session.

Right-clicking and using the "Stop programmatically string sending" feature works for this, but it's manual, annoying, and prone to human error.

This new modifier, when enabled (it's off by default, of course) re-adds the timed event (`Glib::Timeout`), therefore resetting the timer, each time one of the following happens:

- Terminal receives focus
- Terminal cursor changes location (same trigger for changing tab color)
- A character is typed in the terminal

Also includes the extraction of the timer event adder to a separate private function, and the deduplication of one block of code that did the same thing inside the "Stop programmatically string sending" feature.


gfrenoy:
Very cool feature Henrique :)  I was indeed looking for something like this :)

Very minor question : why did you remove the documentation about "Create SOCKS tunnel" / `SOCKS5_PORT` option ([here](https://github.com/asbru-cm/asbru-cm/pull/902/files#diff-f4f46b411f86306cbaa49c13aaff5df931087a59c2f2e0150b1d56132baeea7dL43)) ?  Was this intentional ?

gfrenoy:
How does it look for you ?  Here there is not enough place to see the number of seconds
image

Should we try to make this a bit more compact?

"Programatically send string" --> "Send string"
"With <RETURN>" --> "<RETURN>"
"Every (secs) --> "Every"
"Only when Idle" --> "Idle only"

Also, a specific tooltip explaining whan "Only when Idle" does could be also useful.

That being said it looks to work fine, thanks a lot :)

henry701:
> Very cool feature Henrique :) I was indeed looking for something like this :)
> 
> Very minor question : why did you remove the documentation about "Create SOCKS tunnel" / `SOCKS5_PORT` option ([here](https://github.com/asbru-cm/asbru-cm/pull/902/files#diff-f4f46b411f86306cbaa49c13aaff5df931087a59c2f2e0150b1d56132baeea7dL43)) ? Was this intentional ?

It was intentional because this feature is no longer present on that configuration screen. I think it was moved [here](https://github.com/asbru-cm/asbru-cm/blob/loki/doc/Managing/IncognitoBrowser.md#create-a-random-socks-tunnel) some time go:

image

> How does it look for you ? Here there is not enough place to see the number of seconds image

Here it looks like this, but I'm not using the default font:

image

> Should we try to make this a bit more compact?
> 
> "Programatically send string" --> "Send string" "With " --> "" "Every (secs) --> "Every" "Only when Idle" --> "Idle only"

Will change to Idle only, but I think omitting "secs" from the other label could be bad. Will make it more compact anyways.

> Also, a specific tooltip explaining what "Only when Idle" does could be also useful.

Will do!

henry701:
@gfrenoy Done!

doc/Manual/Connections/SSH.md
```diff
@@ -41,9 +41,9 @@
 ![](images/ssh2.png)
 
 + __Programmatically send a string__ : Send a specified regex expression every selected seconds to the terminal.
-    + _Only when Idle_: Restart the timer (from `Every X seconds field`) each time the terminal is updated or focused.
+    + _Idle Only_: Restart the timer (from `Every X seconds field`) each time the terminal is updated or focused., effectively preventing the string from being sent unless the terminal has been idle for the configured period.
 + __Prepend command__ : Add this command before the ssh command connection string.
-+ __Start next script when connection is launched__ : * Pending
```
henry701, line 46: Oops

PR #1036, opened 2023-05-11
fix: delete script was not working

jojaquix:
A little script deletion fix.
================
Problem: The selected(s) scripts deletion was not working.

Before the change:
----------------------

image
and of course the scripts was not deleted

After the change
-------------------
No error raised and selected script deleted

### DIFF
--- a/lib/PACScripts.pm
+++ b/lib/PACScripts.pm
@@ -1141,7 +1141,7 @@ All $CONNECTIONS{error|out1|out2} are resetted every time a SEND command is exec
         
         # Delete selected files
         foreach my $path (@sel) {
-            my ($file, $name) = $model->get_value($model->get_iter($path), (0));
+            my ($file, $name) = $model->get_value($model->get_iter($path), 0);
             unlink($file);
         }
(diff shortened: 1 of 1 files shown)


jojaquix:
@gfrenoy Do You think this could be merged ? Thanks.

jojaquix:
Ohh great, well I was not able to find perl docs for GTK3 TreeView mode but [here ](https://gtk2-perl.sourceforge.net/doc/pod/Gtk2/TreeModel.html#_tree_model_get_valu) I found the get_value method takes as second parameter a list of indexes and that worked in my tests with current GTK library used by asbru-cm

gfrenoy:
Can you give it a try with "0" if it still works ?

See https://docs.gtk.org/gtk3/vfunc.TreeModel.get_value.html, it looks to expect a "gint".  All the other calls in the same code are also only using an int so let's try to be consistent.

Thanks again for your help !

jojaquix:
I have chaged to use only one gint @gfrenoy and it worked also.

PR #1115, opened 2024-10-10
Add "close other terminals" button

verils:
Like most other Tab-based terminals or text editors, there should be a "close other tabs" button, to close all terminals except for the one I clicked.

I've already achieve this feature, and didn't make too much change. Please have a look at it for review.


gfrenoy:
This one was fairly simple so it's probably OK to have it directly merged to the master branch but if you have the time and motivation to contribute further, better to work on the loki branch ; it will speed up the merge process on our side.

Thanks again for your contribution !

PR #1122, opened 2024-12-20
Add info about forwarding ports to the SSH connection tooltip

efferre79:
I find useful to have the forwarding ports information on the connection tooltip


gfrenoy:
Thanks :)

Issue #640, opened 2020-06-03
Text color with dark themes


beliys:
**The color of text for connections on dark themes is poorly read.**

**To Reproduce**
1. Setup any dark theme for gtk app
2. Run asbru-cm
3. See text color in connections list

**Screenshots**
https://imgur.com/OkszR7x

**Environment (please complete the following information):**
 - OS: Manjaro-KDE latest
 - Ásbrú Version v6.2.1 from AUR (http://aur.archlinux.org/packages/asbru-cm-git)

midniteslice:
I was just coming here with the same issue. 
Version 6.2.0 on kubuntu 20.04 has the same issue. Version 6.1.3 on fedora 31 running kde does not have the issue. 
I have tried many different themes with the same result.

popxunga:
Hi ... 

I do use Ad-waita Dark theme and it looks good. 

My Asbru setup looks like this:

![image](https://user-images.githubusercontent.com/37672088/83652881-06cc4200-a5b3-11ea-805a-3487e301ec38.png)

![image](https://user-images.githubusercontent.com/37672088/83653025-33805980-a5b3-11ea-841d-a06bdab48436.png)

Let us know if that helps.

beliys:
@popxunga , thanks, the "Look and Feel" setting helped me, but it would be better if the default settings were from the system theme.

midniteslice:
That did it for me too. I agree with what beliys said about ti being default. 
Thank You!

popxunga:
Indeed ... the default colors used may need some adjustments ...

gfrenoy:
Not sure what you mean by the "default":

- when you first start Ásbrú (so with a brand new configuration directory)
- when you switch your GTK theme
- when you switch your Ásbrú Icons theme

There is no magic, we need to define the logic that needs to be implemented and I don't see here what to do exactly.

popxunga:
Hi @gfrenoy,
From my point of view, the default text/icon colors should always be relative to the Asbru themes being used, only.
This may be more difficult as it looks on a 1st view, but as basic principle I would say that the logic behind this can be:
For the Asbru dark theme, light colors for text and icons should be used by default;
For the Default or Colored themes, dark colors for text and icons should be used by default.
If the user decides to use the System theme then is not an Asbru problem.

beliys:
I agree with @popxunga , but when using a system theme, the colors should be taken from the theme that the system uses.

popxunga:
[...]

PR #1154, opened 2025-09-02
Dark mode revision

hanspr:
In reference to : #640

I could add a basic validation of the current protected and unprotected colors selected; depending on the selected theme in asbru, and force a safe color for the text (if the contrast is expected to bee to low) so it would be readable in the current theme selected.

The only way to know if a theme is dark or light is based on the system themes name, I use the name to try to select the correct theme mode and have less issues with new users.

Pending: 

asbru.glade

Should replace : "Icon theme" for "Theme"

The css, has minor changes that adjust the colors for dark themes, and now is used ass trigger to have an idea of the correct color settings for the foreground text.

### DIFF
--- a/lib/PACMain.pm
+++ b/lib/PACMain.pm
@@ -4956,8 +4956,8 @@ sub _setSafeLayoutOptions {
         }
     }
     $$self{_CFG}{'defaults'}{'layout previous'} = $layout;
-    # Validate forground colors for corresponging theme
-    
+
+    # Adapt colors of tree nodes if unreadable for currently selected theme
     my $p_set = $$self{_CFG}{defaults}{'protected set'};
     my $p_unset = $$self{_CFG}{defaults}{'unprotected set'} // 'foreground';
     my $p_color = $$self{_CFG}{defaults}{'protected color'};


gfrenoy:
Hi,

Improving the dark mode and "fixing" an obviously unreadable configuration is interesting :)

But about this:
```
    my $mode = Gtk3::Settings::get_default()->get_property("gtk-theme-name");
    if ($$self{_CFG}{'defaults'}{'theme'} !~ /system/i) {
        if ($mode =~ /dark/i && $$self{_CFG}{'defaults'}{'theme'} !~ /dark/i) {
            $$self{_CFG}{'defaults'}{'theme'} = 'asbru-dark';
        } elsif ($mode !~ /dark/i && $$self{_CFG}{'defaults'}{'theme'} =~ /dark/i) {
            $$self{_CFG}{'defaults'}{'theme'} = 'asbru-color';
        }
    }

```

Do we really want to override user's choice ?

While testing, it took me a while to understand why my change to "asbru-dark" was not saved correctly.

At least, we should warn the user, shouldn't we ?

PR #1159, opened 2025-09-17
Click on url

hanspr:
This is a functionality I always wanted to add.

You can position the mouse pointer over a url on the terminal and by control-click on it, will launch the default browser and open that url.

Some times you start a command like `yarn dev` and you get something like this on the terminal

imagen

Or some command spits out a url after an error or warning, or you cat a readme file with urls.

Pending: Change mouse pointer when hovering over link, could not do it, the set_cursor does not work, something changed there on gtk as usual. For the time being is what I can offer, you are more skilled with gtk, maybe you find the way to change the cursor.


hanspr:
fixed, always forget to test tabbed windows.

PR #1160, opened 2025-09-18
Logs to pacterminal

hanspr:
> [!NOTE]
> This only works with vte 0.72+
> On vte 0.72+ will log from PACTerminal, below, will continue logging from asbru_conn
> Added fix from PR #1147, and added extra fix, to disable log if logfile is empty

After all the tinkering with the option to click on urls, I realized that this terminal buffer is completely, clean so it can be used to create the log files.

So in this PR, I condition the log logic from asbru_conn so is replaced by pacterminal if the vte version supports it.

Various improvements with this move.

1. The log files are pristine, no escape sequences, everything in order, simple clean
2. If you name your log file with a suffix : .htm , .html the output of your logfile will be an html log with colors.
3. Find in terminal loaded from the current terminal buffer, does not depends on the logfile
    1. * Find in terminal, old method would pool the info from the log file, but if user configs no log file there is no find in terminal.

* This is the reason why currently asbru produces log files all the time and ignores the flags to create or not a log file. The logfile is used to offer find in terminal. And this conflicts with users requests to avoid logging if not configured.


gfrenoy:
Thanks a lot !  Very clean logs, indeed.

Those changes will be integrated in the next loki version.

Issue #1132, opened 2025-04-29
Using -T (Disable pseudo-terminal allocation) ssh option


Random-Chain:
I am trying to set up a connection that only does local port forwarding from a remote server through a jump host.
The manual command to run is, e.g.: `ssh -L 4321:192.168.1.1:4321 172.16.1.1 -NT`
(when 192.168.1.1 is the remote server and 172.16.1.1 is the jump host)

When trying to set up the same thing in Asbru, I can add the -N option by enabling "Do NOT execute remote commands" in the SSH Options tab. But I can't see any way to add the -T option as well.
What happens is the jump server disconnects after a minute and I can't keep the tunnel open.

For now I have worked around this by creating a Generic Command connection and putting the ssh command as the command to run. 
If there is some way to use -T in SSH connection, I'd love to know what it is.
If there isn't, would be nice if this feature can be added to a future version.

gfrenoy:
Thanks for this proposal !  Change has been integrated into the loki version and will be available in our next main version.

PR #1167, opened 2025-11-03
Option t

hanspr:
Fix: #1132

Add -T option


gfrenoy:
Cool addition.  Thanks !

Issue #1168, opened 2025-11-04
Send enter in one command line


racdamico:
I would like to know if it's possible to send an ENTER in a command line with multiple syntax. For example, create one "single command" like this:
config <send an enter> interface ethernet0/0 <send an enter> show config

is it possible?

hanspr:
#1169

gfrenoy:
Thanks for the proposal and the fix !  This improvement is now available in our "loki" version and will be integrated into the next major release.

PR #1169, opened 2025-11-05
Multiline commands

hanspr:
Allow the use of \n \r \t on command strings to allow multiple line remote/local commands

Long due improvement to allow multiline commands: #1168


gfrenoy:
Thanks !

PR #1170, opened 2025-11-05
Update RemoteCommands.md

hanspr:
document escape sequence substitution proposed on : #1169


gfrenoy:
Always good to have a proper documentation, thanks for that !

PR #1174, opened 2025-11-18
List commands hotkey

hanspr:
In terminal list hotkeys configured in remote commands


gfrenoy:
@hanspr I cannot make this to work...

If the hotkey has never been used before, nothing is displayed.
If I first use the hotkey, then I see it in the list but there is no label.

I'm not 100% sure but I believe the way we loop into the $kbs variable is not correct.

I'll give another look tomorrow.

hanspr:
That is shown up on using it first it is a weird behavior, I will try to replicate during the week.

Regarding the missing label, I assume that there is a description in your configuration of the remote command (I ask because have seen from previous posts from users that many people tend to leave this blank), and the label comes from the description.

imagen

gfrenoy:
> I assume that there is a description in your configuration

Yes, there is one.

Will try again tomorrow and give more details.

hanspr:
I started asbru with a new clean configuration: `./asbru-cm --config-dir=/home/hans/.config/asbru-new --verbose`

Created a new remote command

imagen

Opened a local terminal, and tested, was shown correctly without using it, or anything

imagen

Closed the terminal, Added a second remote command

imagen

Opened a new local terminal. and tested

imagen

---

So, removed the clean configuration and recreated again from scratch, and this time I opened the terminal first. Then created the hotkey for the remote command, and then checked if it was shown, and it worked too.

imagen

imagen

So at least, from a new configuration it worked, so it means the problem should be around pre existing configurations. I will test that next.

gfrenoy:
Ok I see the « problem » !  I tested this with **connection** remote commands.  Not the global ones !

So for the global commands, it works fine.  But if you define a command at the connection level, it behaves like described above.  I guess this is because there is one level more in the structure.

Do we want to support the connection commands as well ?

gfrenoy:
This is how the data looks like when you create connection-specific remote (or local) commands:

```
{
          '9b42d0c7-57c1-46dd-9093-43f857a99d0f' => {
                                                      'terminal' => {
                                                                      'Ctrl+l' => [
                                                                                    'User hotkey',
                                                                                    'HOTKEY_CMD:List Files Remote',
                                                                                    'ls -l
'
                                                                                  ],
                                                                      'AltCtrl+l' => [
                                                                                       'User hotkey',
                                                                                       'HOTKEY_CMD:List Files Local',
                                                                                       'ls -l
'
                                                                                     ]
                                                                    }
                                                    },
          'terminal' => {
                          'AltCtrl+k' => [
                                           'User hotkey',
                                           'HOTKEY_CMD:List Files Remote Global',
                                           'ls -l
'
                                         ]
                        }
        };
```

The first part is ignored.

gfrenoy:
Made a few minor changes to handle the various cases.  It works fine, thanks a lot for this cool improvement :)

PR #1187, opened 2026-04-12
OSC 52 clipboard support — lets tmux and remote programs write to the local clipboard

### LINKED ISSUE #631: Copy from tmux buffer to system buffer (clipboard)
hi, i have a problem copying from Ásbrú with tmux, to my system clipboard .. it only happens with Ásbrú, because for example with Konsole if you copy to the system clipboard ..

I leave some tmux settings.

https://github.com/gpakosz/.tmux

 - OS: Archlinux / Plasma
 - Version : loki  2fe4c2f

davidgut1982:
## Summary

Closes #631 — *Copy from tmux buffer to system buffer (clipboard)*.

This PR implements OSC 52 escape-sequence clipboard support using a novel architectural approach that requires **no changes to VTE** (which explicitly refuses OSC 52 — GNOME bug #795774). A passive `Expect::log_file` coderef observer is installed in `asbru_conn` that scans raw terminal output for OSC 52 sequences and dispatches them to `PACTerminal` via the existing `ctrl()` Unix-socket channel.

To our knowledge this is the **first OSC 52 implementation in any Perl+GTK3 terminal**.

## How it works

```
asbru_conn (child process)
  └─ Expect::log_file coderef  ← passive observer, fires on ALL output
       └─ _osc52_scan_chunk    ← scans for ESC]52;...
            └─ ctrl("OSC52:c:<base64>")  ← existing Unix socket
                 └─ PACTerminal._watchConnectionData
                      └─ _handleOsc52Payload  ← GTK3 clipboard write
```

## Files changed

| File | What changed |
|---|---|
| `lib/asbru_conn` | `Expect::log_file` coderef wrapper + `_osc52_scan_chunk` scanner |
| `lib/PACTerminal.pm` | `_handleOsc52Payload` handler + kill-switch env var propagation |
| `lib/PACUtils.pm` | Config defaults (`allow osc52 write = 0`, opt-in) |
| `lib/PACConfig.pm` | Load/save wiring for new Preferences checkbox |
| `res/asbru.glade` | `cbCfgAllowOsc52Write` GtkCheckButton in General tab |
| `NEWS.md` | Feature documentation |

## Security model

- **Opt-in by default** — disabled until the user enables it in Preferences → General
- **Kill-switch** — `ASBRU_OSC52_DISABLE=1` env var overrides config and disables entirely
- **Size cap** — sequences larger than 100 KB are silently dropped
- **Base64 alphabet validation** — strict; also blocks ctrl-channel injection attacks
- **Query form dropped** — `52;c;?` (read-clipboard) is never honoured
- **16 MiB buffer cap** — prevents DoS via unterminated sequences

## Verified working (Manjaro Linux, VTE 0.82, tmux 3.6a)

| Scenario | Result |
|---|---|
| Direct `printf` OSC 52 in PACShell | ✅ |
| Direct `printf` in SSH session | ✅ |
| tmux mouse-select → clipboard | ✅ |
| OSC 52 with empty targets field (tmux default per spec) | ✅ |
| UTF-8 / Unicode content (box-drawing, emoji, accented letters) | ✅ |
| Kill-switch `ASBRU_OSC52_DISABLE=1` | ✅ |
| Size cap enforcement | ✅ |

## Notable implementation details

Two spec-compliance bugs were found during real-world testing that static review missed:

1. **Empty targets field** — per the xterm `ctlseqs` spec, an empty `Pc` field means "use default" (CLIPBOARD). tmux sends empty targets by default for mouse selections. The scanner now normalises empty targets to `'c'` instead of dropping the sequence.

2. **UTF-8 flag on clipboard write** — `MIME::Base64::decode_base64()` returns bytes with Perl's UTF8 flag unset. Passing this to `Gtk3::Clipboard::set_text()` causes the GLib binding to re-encode as Latin-1, producing mojibake for non-ASCII text. Fixed with `Encode::decode('UTF-8', $bytes)` before the clipboard write.

## Debug logging

`ASBRU_OSC52_DEBUG=1` logs every matched sequence to `/tmp/asbru_osc52_debug.log` — permanently available for user support without source patches.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,23 @@
 
 Here you can find important news on the project
 
+## 12.04.2026
+
+- **OSC 52 clipboard support** — Remote programs running inside Ásbrú sessions (tmux
+  mouse-select, Neovim, Helix, etc.) can now write to the local system clipboard via
+  [OSC 52 escape sequences](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands).
+  This closes the long-standing [issue #631](https://github.com/asbru-cm/asbru-cm/issues/631).
+
+  **How to enable:** Preferences → General → *Allow OSC 52 clipboard writes (tmux / remote terminals)*.
+  The feature is **opt-in** (disabled by default) for security — enable it only when connecting
+  to hosts you trust.
+
+  **Kill-switch:** Set the environment variable `ASBRU_OSC52_DISABLE=1` before launching
+  Ásbrú to disable OSC 52 processing entirely, regardless of the Preferences setting.
+
+  **Debug logging:** Set `ASBRU_OSC52_DEBUG=1` to log every matched OSC 52 sequence to
+  `/tmp/asbru_osc52_debug.log` — useful when reporting issues.
+
 ## 04.04.2026
 
 After almost 4 years we are happy to release version 6.4.1 of Ásbrú Connection Manager.
--- a/lib/PACConfig.pm
+++ b/lib/PACConfig.pm
@@ -863,6 +863,7 @@ sub _updateGUIPreferences {
     _($self, 'rbOnNoTabsClose')->set_active($$cfg{'defaults'}{'when no more tabs'} == 1);
     _($self, 'rbOnNoTabsHide')->set_active($$cfg{'defaults'}{'when no more tabs'} == 2);
     _($self, 'cbCfgSelectionToClipboard')->set_active($$cfg{'defaults'}{'selection to clipboard'});
+    _($self, 'cbCfgAllowOsc52Write')->set_active($$cfg{'defaults'}{'allow osc52 write'});
     _($self, 'cbCfgRemoveCtrlCharsConf')->set_active($$cfg{'defaults'}{'remove control chars'});
     _($self, 'cbCfgLogTimestam')->set_active($$cfg{'defaults'}{'log timestamp'});
     _($self, 'cbCfgAllowMoreInstances')->set_active($$cfg{'defaults'}{'allow more instances'});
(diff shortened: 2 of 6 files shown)


gfrenoy:
Thanks for the proposal.

But beyond the technical details, can you explain what practical use case it will help to solve ?  I mean: as a user, what will it bring to me ?  I can already do copy/paster from a tmux session in Ásbrú, I'm not sure to understand the need... (the referred issue has been closed for years).

davidgut1982:
Thanks for asking — it's a fair question and worth explaining clearly.

You're right that Shift+click already works for selecting visible text in Asbru. The gap OSC 52 fills is different: it's for when a remote application — not the user's mouse — decides what goes on the clipboard.

The two scenarios that come up constantly in practice:

1. Vim/Neovim yank over SSH. When you press yy or y} in vim on a remote server, vim knows exactly what it yanked — including precise whitespace, trailing newlines, etc. With OSC 52, that lands directly in your local clipboard. Without it, you have to Shift+drag the text visually in the terminal, which may not match what vim actually copied (indentation gets clipped, wrong line boundaries, etc.). Plugins like vim-oscyank and nvim-osc52 have become standard in remote dev setups specifically because this is so common.

2. tmux keyboard copy-mode. When a user does Ctrl+B [ and selects text with the keyboard (common in headless/no-mouse workflows), tmux sends OSC 52 to write to the system clipboard. Without OSC 52 support in the terminal, that selection stays in tmux's internal buffer only — it never reaches the system clipboard, so you can't paste it in another application. Shift+click doesn't help here because the user is inside tmux's copy-mode, not making a normal terminal selection.

There's also a third case that's become more relevant recently: tools like lazygit, fzf, and various shell scripts emit OSC 52 directly to copy output (file paths, commit hashes, etc.) to clipboard without the user selecting anything.

In all of these, the common thread is: the remote application is the one initiating the copy, not the user's mouse. Shift+click can't cover that.

OSC 52 is now supported by essentially every modern terminal (Kitty, Alacritty, WezTerm, Windows Terminal, Konsole since 2022). For SSH + tmux + Neovim users — which is an increasingly common workflow — Asbru being the one terminal that doesn't support it is a real friction point.

gfrenoy:
The reason why it has been rejected (so far...) by the Gnome project are interesting to read.
https://gitlab.gnome.org/GNOME/vte/-/issues/2495

At this stage, I would tend to rely on what VTE is doing.  I don't see this as something Ásbrú should manage differently.

[Cheers to [Egmont](https://github.com/egmontkob), btw, who significantly contributed to this project as well]

davidgut1982:
Thanks for pointing to that issue — it's worth addressing directly.

VTE's refusal makes sense for a general-purpose library: they can't control the security context of every application that embeds them, so a conservative default is reasonable.

The architecture here is different though. asbru_conn runs as VTE's PTY child and uses a passive observer on the SSH subprocess output to detect OSC 52 sequences. VTE does receive the raw bytes, but since VTE doesn't implement OSC 52 it silently discards them — they don't appear on screen. Asbru separately dispatches the payload via its own Unix socket channel to handle the clipboard write directly in PACTerminal, without involving VTE's clipboard APIs at all. VTE's decision not to implement this stays intact.

The security concerns VTE raised are addressed: the feature is opt-in (disabled by default), there's a 100 KB size cap, strict base64 validation, and a kill-switch env var. Same write-only threat model Konsole adopted in 2022.

